PR #1, opened 2022-07-27
Go support

neekolas:
## Summary
- Initial commit on the repo. Adds support for generating Go files
- Adds some hooks to automatically generate the Go files on every commit
- Drops in the GRPC API proto files from xmtp-node-go
- Github actions for linting

## Coming up
- Typescript codegen

## Notes
There is a blocker right now in Typescript. When generating the files, there is an invalid import in the generated code:
```
import * as GoogleProtobufTimestamp from "./google/protobuf/timestamp.pb"
```
https://github.com/xmtp/xmtp-js/compare/main...message-grpc#diff-cabf66a344e09eb97d00ca7b71b805b4d9b667b303e0edb23c4d868cdca8e57aR8

That file does not exist. Looking into how to point it to the real location of the Google type.


.github/workflows/lint.yml
```diff
@@ -0,0 +1,22 @@
...
+    steps:
+      - uses: actions/setup-go@v2
+        with:
+          go-version: "1.17.8" # The Go version to download (if necessary) and use.
+      - uses: actions/checkout@v2
+      - name: golangci-lint
+        uses: golangci/golangci-lint-action@v3
+        with:
+          version: v1.45.2
+          args: --timeout=5m
```
snormore, line 22: Can we add a [protolint](https://github.com/marketplace/actions/protolint) job here too? And maybe a `dev/lint` script that calls `protolint` with `-fix`?

dev/go/vendor-proto
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env sh
+set -e
+
+vendor_dir="build/tmp/vendor"
```
snormore, line 4: I think we'll need/want something like this in `.vscode/settings.json`:
```json
{
    "protoc": {
        "compile_on_save": false,
        "options": [
            "--proto_path=build/tmp/vendor",
        ]
    }
}
```

.github/workflows/lint.yml
neekolas: Done

.protolint.yaml
```diff
@@ -0,0 +1,149 @@
+---
```
snormore, line 1: We can probably remove this file and just use the default (no `.protolint.yaml`) to start - unless there were specific rules in here you had changed

snormore (approved):

.protolint.yaml
neekolas: I did change some rules. They really don't like the single word RPC names, and they require a comment on every enum field. Both of which I find annoying.

I also temporarily disabled the requirement for a comment on every field, which I would like to turn back on once the dust settles on the definitions.

mkobetic:
Looks good, to so the go import path should work out to `github.com/xmtp/proto/go/message_api/v1` right? That looks reasonable.

mkobetic:
I wonder if `proto` is a bit too general though, since this is aimed at client <-> network protocol at this point.

neekolas:
> I wonder if `proto` is a bit too general though, since this is aimed at client <-> network protocol at this point.

I thought about that. But I'm hoping we will start to move some of our client protos into this repo as well. Will be particularly important once we have multiple clients (mobile)

PR #2, opened 2022-07-27
Update proto

neekolas:
## Summary
Makes some updates to the protos to closer to feature parity with the existing LibP2P APIs. Notably
- Adds startTime and endTime fields to the `QueryRequest`
- Adds simplified pagination to the `QueryRequest`
- Adds a timestamp to the `PublishRequest`
- Moves `SortDirection` into a top level message so it can be used in multiple places
- Only allows publishing to one content topic at a time (seems like multi-publish is a potential DOS vector, and doesn't serve have much utility for our use-cases)

## Notes
I made the cursor a `oneof` type, since I would really like to change the way we handle cursor based pagination to have a simpler query structure. The pagination queries are needlessly complex, and would ideally be based off of a single field. I initially adopted this model to be more consistent with `go-waku`, who then changed their structure to a better model after we shipped.


snormore (approved):

mkobetic (approved):
Excellent!

PR #3, opened 2022-07-27
Force dependency versions

neekolas:
## Summary
Use some blank imports to keep dependency versions pinned in the `go.mod`


neekolas:
Linter doesn't like that very much, so I'll try and shut it up. 
CleanShot 2022-07-27 at 11 54 27@2x

mkobetic (approved):
❤️

mkobetic:
This seems to be exactly what the docs prescribe https://github.com/grpc-ecosystem/grpc-gateway#installation

PR #4, opened 2022-07-27
Typescript Support

neekolas:
## Summary
- Includes some build scripts for Typescript proto generation, borrowed and adapted from @snormore's previous work on the client SDK
- Some prototype-y Typescript compilation. Definitely needs some testing to ensure this package is actually importable
- Changes all `google.protobuf.Timestamp` objects to `uint64`. [See this conversation](https://xmtp-labs.slack.com/archives/C02BABHFZG9/p1658961552938639). Technically message timestamps can be negative in Waku. I think we should just validate against that so we can assume all timestamps are positive, but I could switch to `sint64` to avoid that validation if we really care.
- Setup `semantic-release` to publish this as a NPM package

## Next Steps
- Ensure JS code actually runs in different environments


neekolas:
One thing that's pretty cringey about the generated code with `protoc-gen-grpc-gateway` is that the type for a `uint64` is actually a `string`. Guess that's how they send it over the wire, and there is no runtime transformation of numeric types to strings for you.

CleanShot 2022-07-27 at 16 06 41@2x

mkobetic:
Didn't you want to commit the generated code?

mkobetic:
> One thing that's pretty cringey about the generated code with `protoc-gen-grpc-gateway` is that the type for a `uint64` is actually a `string`. Guess that's how they send it over the wire, and there is no runtime transformation of numeric types to strings for you.
> 
> CleanShot 2022-07-27 at 16 06 41@2x

Uff, that is crappy, anything reasonable alternative that would come out better?

dev/ts/generate
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env bash
+set -e
+
+(
+  cd dev/ts/protoc
+  docker build -t xmtp/protoc .
```
mkobetic, line 6: We don't use docker for the Go side, does this make it easier?

neekolas:
> Didn't you want to commit the generated code?

Generally I hate committing generated code. It leads to noisy and hard to read diffs and new places to find merge conflicts.

Go just doesn't leave you much choice in the matter. 

But for TS it's both easy and an established pattern to generate everything on demand before you publish.

dev/ts/generate
neekolas: I _could_ use it for both. The go side runs the generator in a pre-recommit hook, so I'm trying to keep that flow as lean as possible. Given how rarely this changes, was also thinking of breaking this out into a tiny repo of its own so that we don't have to build the image on its own.

neekolas:
> Uff, that is crappy, any reasonable alternative that would come out better?

Haven't found anything that lets me get better types out of the generator. What I am working on is a more robust `ApiClient` that wraps the generated code in the SDK, and adds some better types as well as handles things like pagination.

mkobetic (approved):

mkobetic:
I'll update xmtp-node-go for the new timestamp format.

PR #5, opened 2022-08-04
Use Envelope in PublishRequest

mkobetic:
Writing some publish tests I noticed that I need to [transform Envelope into PublishRequest](https://github.com/xmtp/xmtp-node-go/blob/main/pkg/api/server_test.go#L132-L136), which seems kinda pointless and arguably inconsistent with the rest of the API which returns Envelopes. We are also losing Envelope.Id which is probably not good either. So I'm proposing to just wrap the Envelope in the PublishRequest instead (do we want to allow collection of them maybe?).


go/message_api/v1/message_api_grpc.pb.go
```diff
@@ -1,7 +1,7 @@
 // Code generated by protoc-gen-go-grpc. DO NOT EDIT.
 // versions:
 // - protoc-gen-go-grpc v1.2.0
-// - protoc             v3.21.2
+// - protoc             v3.19.4
```
neekolas, line 4: You'll want to upgrade your `protoc` version to make these go away

neekolas:
I wouldn't be against taking a collection of Envelopes. Seems like there are plenty of valid cases for publishing multiple messages at once that we shouldn't be blocking at the API level.

neekolas:
This is entirely my fault, since I haven't yet written a README for the `proto` package, or setup the commitlint. But you'll want to use the same "semantic-release" format for commit messages to trigger new releases for the JS SDK.

I'll fix those issues today.

neekolas:
For publishing, we would expect `Envelope.id` to be blank, right? Those get generated server side right now, and I'd rather not reproduce the logic in case we get inconsistencies.

mkobetic:
> This is entirely my fault, since I haven't yet written a README for the `proto` package, or setup the commitlint. But you'll want to use the same "semantic-release" format for commit messages to trigger new releases for the JS SDK.

I'll squash-rebase to clean up the commit, what should the commit message be in this case? This is a breaking change so should we roll the major number?

neekolas:
Given that the library isn't being used in `main` of the SDK yet, I think the number doesn't matter much as long as it's incremented at all. But in theory it should be a breaking change.

mkobetic:
> For publishing, we would expect `Envelope.id` to be blank, right? Those get generated server side right now, and I'd rather not reproduce the logic in case we get inconsistencies.

Hm, good point. Is this an argument against this change then?

neekolas:
Clients do care about the ID for read APIs (for example, to dedupe messages). Proto fields are optional, so we could just not use it on Publish. If we were to ship this change, I would just make the API we expose to users exclude the field and leave it blank in the API calls.

mkobetic:
Hm, looking at the node side it seems the ID is [mixed up with the pagingInfo.Cursor.Digest](https://github.com/xmtp/xmtp-node-go/blob/main/pkg/store/db_store.go#L87). Isn't that field sufficient for the querying side of things? You wouldn't get it when you stream messages via Subscribe, but you arguably shouldn't need to dedupe in that case, no?

neekolas:
We also want `ID` on the messages themselves. For example, in the example chat app we dedupe messages between streaming and list API calls using the ID. It's also very helpful in a React context for the `key` field.

mkobetic:
I'm missing parts of the puzzle here, I don't see a way to actually get the stored Id. Relayed messages don't have it at all and HistoryResponse also just has []*WakuMessages and those don't have Id either. So how are we getting it at all?

neekolas:
Maybe it's actually being generated on the client right now via `js-waku` (which we no longer have)

mkobetic:
Looks like it's just generated on the client side https://github.com/xmtp/xmtp-js/blob/main/src/Message.ts#L70-L71. So we should be able to just drop it here right?

neekolas:
In any case, it's going to be challenging to generate consistent IDs on the client now, given that we serialize things so differently.

mkobetic:
> In any case, it's going to be challenging to generate consistent IDs on the client now, given that we serialize things so differently.

It seems it's generated just from our proto payload so that should remain consistent.

mkobetic:
OK, rebased, cleaned up, commit message fixed, GTG?

neekolas:
I guess that means our IDs have always been different from server IDs

neekolas (approved):

mkobetic:
> I guess that means our IDs have always been different from server IDs

I'm not sure we really have server IDs, they seem to only exist in store.Envelope and get throw away immediately after retrieval from DB. So they don't seem to exist outside of the store subsystem.

neekolas:
LibP2P gossipsub definitely cares about IDs as well. It just calls it `Hash` in that context.

PR #6, opened 2022-08-05
Add fetch polyfill

neekolas:
## Summary
- Adds a polyfill for the `fetch` API to this repo.
- Updates the README to explain the precommit hooks and how to setup development

## Notes
- All of the generated proto code relies on the Fetch API, which is standard in browsers but not in Node. This PR creates a fetch polyfill using [Undici](https://undici.nodejs.org/#/), which adheres closer to the fetch WHATWG standard than other fetch polyfills you can find on NPM. I tried the more standard fetch polyfill (cross-fetch), but it conflicts with @stardazed/streams-polyfill - which we use for message compression - and causes the application to crash on startup. This appears to be because of the non-standard body types [explained here](https://github.com/node-fetch/node-fetch/blob/main/docs/v3-LIMITS.md). Given the focus on standards-compliance, and the support of Node.js core (it is in the process of merging into Node.js itself), it feels like a better choice than cross-fetch even if we have to do the polyfilling ourselves.
- We use the `browser` field of the package.json to not include the polyfill module in browser bundles made with Rollup or Webpack.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,2 +1,11 @@
 # proto
+
 Shared Protocol Buffers and their associated generated code
+
+## Working on this repo
+
+Before committing any changes to this repo, you will want to run `npm i` at least once. This will install the Husky precommit hooks to your machine.
+
+Go code will be regenerated based on the `.proto` files on each commit and does not need to be done manually.
+
+Commit messages must be in the [Angular Commit Message Convention](https://gist.github.com/stephenparish/9941e89d80e2bc58a153)
--- a/package.json
+++ b/package.json
@@ -2,12 +2,13 @@
   "name": "@xmtp/proto",
   "version": "0.0.0-development",
   "description": "Protobuf client and generated classes for GRPC API",
-  "main": "./ts/dist/cjs/index.js",
-  "module": "./ts/dist/esm/index.js",
+  "main": "./ts/dist/cjs/node.js",
+  "module": "./ts/dist/esm/node.js",
   "types": "./ts/dist/types/index.d.ts",
+  "browser": "./ts/dist/esm/index.js",
   "exports": {
     ".": {
-      "types": "./ts/types/cjs/index.d.ts",
+      "types": "./ts/types/types/index.d.ts",
       "import": "./ts/dist/esm/index.js",
       "require": "./ts/dist/cjs/index.js"
     }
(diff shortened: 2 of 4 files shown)


package.json
```diff
@@ -2,12 +2,13 @@
   "name": "@xmtp/proto",
   "version": "0.0.0-development",
   "description": "Protobuf client and generated classes for GRPC API",
-  "main": "./ts/dist/cjs/index.js",
-  "module": "./ts/dist/esm/index.js",
+  "main": "./ts/dist/cjs/node.js",
+  "module": "./ts/dist/esm/node.js",
   "types": "./ts/dist/types/index.d.ts",
+  "browser": "./ts/dist/esm/index.js",
```
neekolas, line 8: This is how I conditionally exclude the polyfill package entirely from browser builds

elisealix22: Is there an advantage to using the browser-provided fetch? I would almost lean to using the polyfill for everything for consistency / making it easier to debug since we have to support it anyways

README.md
```diff
@@ -1,2 +1,11 @@
...
+
 Shared Protocol Buffers and their associated generated code
+
+## Working on this repo
+
+Before committing any changes to this repo, you will want to run `npm i` at least once. This will install the Husky precommit hooks to your machine.
+
+Go code will be regenerated based on the `.proto` files on each commit and does not need to be done manually.
+
+Commit messages must be in the [Angular Commit Message Convention](https://gist.github.com/stephenparish/9941e89d80e2bc58a153)
```
elisealix22, line 11: Out of curiosity: Is this used to generate something? Or just a preference for the repo?

elisealix22 (commented):
Love the `Working on this repo` section! Just a couple of questions for you.

package.json
neekolas: Undici only works in browsers with some complicated WASM builds, since it's full of native code. It'll also blow up the bundle size (it's quite large).

README.md
neekolas: It gets used to version the NPM package, just like XMTP JS

elisealix22 (approved):

README.md
elisealix22: Nice! Might be useful to add that note to the README for clarity on how it's used, but it could be declared somewhere else and I'm not seeing it.

PR #7, opened 2022-08-09
Add authn protocol

mkobetic:
This PR proposes a fairly significant change to our authn story. Instead of running a separate protocol, we would just add a `Token` type, carrying pretty much the same payload as `authn.Request` does, to the `message/v1` protocol. The Token  would need to be attached to every request and validated by the nodes (again same type of validation as authn did), so the validation would happen on every request instead of on every run of the authn protocol (although significant chunk of this could be cached since the token remains the same during its validity period).

The token can be carried as a standard HTTP/GRPC authorization header (The gateway maps these automatically https://github.com/grpc-ecosystem/grpc-gateway#mapping-grpc-to-http). The server side processing can be implemented with a GRPC interceptor.

Upsides:
* no need to orchestrate the interaction of the the two protocols
* the token is portable across nodes, so the client doesn't need to "reauthenticate" if it gets loadbalanced to a different node
* implementation should be simpler

Downsides:
* The token/signature validation happens on every request (we can cache some of that if needed)
* We are losing the channel binding of the token that was mediated by the peer IDs, so the blast radius of a leaked token is necessarily larger (but comparable to JWT token style authorization)


proto/authn/v1/authn.proto
```diff
@@ -0,0 +1,70 @@
...
+  
+  message ClientAuthResponse{
+    oneof version {
+      V1ClientAuthResponse v1 = 1;
+    }
+  }
+  
+  message AuthData {
+    string wallet_addr = 1;
+    string peer_id = 2;
```
neekolas, line 68: It seems like we are going to need to remove the `peer_id` and figure out some sort of alternative identifier

mkobetic:
@neekolas I'm not entirely sure what to do on the TS side. It doesn't seem like I should just add
```
export * from "./auth/v1/authn.pb"
```
to `ts/index.ts`. Wouldn't that yield a single package with everything mixed together?

neekolas:
> Wouldn't that yield a single package with everything mixed together?

Ya, we should probably move to namespacing the exports. `export * as authn from './auth/v1/authn.pb'`

neekolas:
Although we don't have to export anything right now. All we are ultimately going to care about from Typescript is whatever GRPC Gateway API gets exposed. We shouldn't need any typescript for the actual protos.

mkobetic:
Hrm, this will be a problem, whan I run `npm run generate` I get:
```
authn/v1/authn.proto: is a proto3 file that contains optional fields, but code generator protoc-gen-grpc-gateway-ts hasn't been updated to support optional fields in proto3. Please ask the owner of this code generator to support proto3 optional.--grpc-gateway-ts_out: 
```

neekolas:
@mkobetic There is a hack listed in [this issue](https://github.com/grpc-ecosystem/protoc-gen-grpc-gateway-ts/issues/21), but it's not great. The suggestion is just to modify the generator and make it register itself as supporting optional fields when it doesn't. Given that we don't actually have optional fields (I'm not entirely sure why this is triggered, but I suspect it's because `oneof` fields with a single option are effectively the same as `optional` in proto3) that may be enough to make it generate valid code.

Open to better suggestions.

proto/authn/v1/authn.proto
```diff
@@ -0,0 +1,71 @@
...
+message ClientAuthRequest {
+    PublicKey identity_key = 1;
+    bytes auth_data_bytes = 3;
+    Signature auth_data_signature = 4;
+}
+  
+message ClientAuthResponse {
+    oneof union {
+        bytes token = 1;  // if auth was successful
+        string error_str = 2; // if auth was not successful
```
neekolas, line 53: Why can't we just use GRPC's native status codes, with custom messages, to pass error responses?

mkobetic:
> @mkobetic There is a hack listed in [this issue](https://github.com/grpc-ecosystem/protoc-gen-grpc-gateway-ts/issues/21), but it's not great. The suggestion is just to modify the generator and make it register itself as supporting optional fields when it doesn't. Given that we don't actually have optional fields (I'm not entirely sure why this is triggered, but I suspect it's because `oneof` fields with a single option are effectively the same as `optional` in proto3) that may be enough to make it generate valid code.
> 
> Open to better suggestions.

It's because of the optional Signature on the PublicKey. We can actually remove optional on it since we require the wallet signature anyway. That's also why I dropped the separate signature field from the message, we my as well use the field on the PublicKey struct (I understand that Jazz was trying to preserve the original definition, but I don't think we really need to in this case). So I just removed optional on that field and it's not complaining now.

neekolas:
OK, great. `optional` fields are kinda silly in protobuf since all fields are technically optional

proto/authn/v1/authn.proto
mkobetic: Maybe we can, I'm not quite sure how that works. I'm still just playing with this, trying to figure out what I want to do.

At this point I'm thinking the token could really be just a random number, that we associated with the wallet when auth succeeds (this is the replacement for peer Id in peer store). The client will have to pass the token along with every request to the API and until the token is purged from the store (purgeLoop) the requests will be allowed. I think this would be an equivalent of what we had.

But then it gets interesting what happens when we purge the token and the connection is still active, somehow the client would have to re-engage the auth flow to get a new token. Makes me wonder if we couldn't just use the AuthData bundled as a token with every request to begin with. Then we don't have to worry about the protocol interplay and requests will be automatically portable between nodes too. Thoughts?

neekolas:
@mkobetic OK, great.

proto/authn/v1/authn.proto
mkobetic: To clarify the token would be produced by the client and it would sign a timestamp and wallet address using the identity key signed by the wallet. The server would just check the token on every request. If the token is too old or signatures don't check out, the request is rejected (probably using GRPC interceptors to implement this). That's it.

proto/authn/v1/authn.proto
```diff
@@ -0,0 +1,71 @@
...
+    string wallet_addr = 1;
+    uint64 timestamp = 3;
+}
+
+////////////////////////////
+
+service AuthnApi {
+    rpc Authenticate (ClientAuthRequest) returns (ClientAuthResponse) {
+        option (google.api.http) = {
+        post: "/authn/v1/authenticate"
```
neekolas, line 67: Feels like this and the next two lines should be indented. Does the formatter let you?

mkobetic: I just run `dev/lint` on this and it mostly autofixes things. I'll check.

proto/authn/v1/authn.proto
neekolas: That doesn't sound crazy to me. Little bit of extra overhead verifying all the signatures on each publish, but there's no network calls or anything like that. I like that it's stateless and portable.

Apps would be able to generate tokens infinitely into the future, but that was possible with the previous flow as well (so long as they kept the same peerID). Probably not our biggest worry. 

Wdyt @jazzz?

mkobetic: We can reject tokens with future timestamps.

proto/authn/v1/authn.proto
```diff
@@ -0,0 +1,71 @@
...
+}
+  
+message ClientAuthResponse {
+    oneof union {
+        bytes token = 1;  // if auth was successful
+        string error_str = 2; // if auth was not successful
+    }
+}
+  
+message AuthData {
```
neekolas, line 57: One thing we are going to have to figure out is how we consistently serialize this `AuthData` into `auth_data_bytes` between the client and server, now that we only have a JSON API and no client-side protobuf. Whatever gets serialized into `auth_data_bytes` needs to be verifiable by the server. I guess we could use actual-protobuf just for this step, but that feels awkward.

proto/authn/v1/authn.proto
mkobetic: Also, in general signature verification should be cheaper than signing (unless that changed with the EC world). So the important thing performance wise would probably be caching the token on the client side while it's valid.

neekolas: It's a case of stockpiling. You could generate them to the year 3000 and then use them at the right moment.

neekolas: Will definitely cache in the client. We'll need to agree on some validity window (60 mins?) and regenerate if expired.

mkobetic: Looks like I was wrong https://www.ietf.org/proceedings/92/slides/slides-92-lwig-3.pdf, claims that EC verification is slower than EC signing :til:. Even so, if it becomes an issue, nodes can cache tokens that they've already seen as well, if it comes to that.

neekolas: We would have to make sure this whole thing serializes into a small enough size to fit in the HTTP Headers. Usually that gets cut off at 4kb or 8kb

mkobetic: > It's a case of stockpiling. You could generate them to the year 3000 and then use them at the right moment.

Interesting, I'm not sure what would motivate a client to do that. It just increases a chance of someone stealing the token and impersonate them. I think rejecting future tokens can help making far future tokens less valuable. They would still have an "expiration window" sized period where it would be valid.

neekolas: > I'm not sure what would motivate a client to do that

Ya, most of the cases I can think of would be in the event of a malicious client...in which case we have bigger problems to worry about, since they can generate new tokens whenever they want.

proto/authn/v1/authn.proto
mkobetic: I think protobuf makes sense since we are using protobuf for messages themselves.

proto/authn/v1/authn.proto
mkobetic: JWT auth with GRPC interceptors https://dev.to/techschoolguru/use-grpc-interceptor-for-authorization-with-jwt-1c5h

proto/authn/v1/authn.proto
neekolas: I'm fine with using `protobuf`. Keeps things leaner, which matters when you are stuffing it into a HTTP header.

Wouldn't you have to include the full `ClientAuthRequest` into the header to do the real validation and tie it to a wallet address? Not just the `AuthData`, but the pub key and wallet signature.

proto/authn/v1/authn.proto
mkobetic: > We would have to make sure this whole thing serializes into a small enough size to fit in the HTTP Headers. Usually that gets cut off at 4kb or 8kb

Yup, shouldn't be too bad, the signatures should be on the order of the key sizes, so I'd estimate good bit less than 1kb total.

proto/authn/v1/authn.proto
mkobetic: Sorry, yes, the idenity key and the signatures would have to be part of it.

neekolas: If we are going with this suggestion (which I'm into assuming Jazz doesn't have any concerns), then we would want to do the codegen here with `protoc-gen-ts_proto` just like we do in the `buf.gen.yaml` on the client.

mkobetic:
@neekolas, the latest is how it would look if Authn was just a Token that is part of the message/v1 protocol. WDYT?

dev/ts/generate
```diff
@@ -7,4 +7,4 @@ set -e
 )
 
 rm -f ./ts/**/*.pb.ts
-docker run --rm -i -v${PWD}:/code xmtp/protoc -I=./proto --grpc-gateway-ts_out=ts message_api/v1/message_api.proto
\ No newline at end of file
+docker run --rm -i -v${PWD}:/code xmtp/protoc -I=./proto --grpc-gateway-ts_out=ts message_api/v1/message_api.proto message_api/v1/authn.proto
```
neekolas, line 10: We are going to want to change this to use `protoc-gen-ts_proto` for the `authn` protos to get real protobuf generated code. This is going to generate types and code for a GRPC Gateway JSON API.

Think we'll need to add the `--ts_proto_opt=esModuleInterop=true` flag as well

proto/message_api/v1/authn.proto
```diff
@@ -0,0 +1,66 @@
...
+    oneof union {
+        Secp256k1Uncompresed secp256k1_uncompressed = 3;
+    }
+}
+
+///////////////////////////////////////////////////////////////////////////////
+
+// Token is used by clients to prove to the nodes
+// that they are serving a specific wallet.
+message Token {
```
neekolas, line 46: This looks good

dev/ts/generate
mkobetic: Can you suggest the change you have in mind. This stuff is black magic to me.

neekolas: I'll try and commit it to this branch. It'll take a little fiddling for me to figure out too. Protoc arguments never work the way I expect them to.

proto/authn/v1/authn.proto
jazzz: Few Questions:

- In this scheme how is the node providing the sybil resistance portion of Authn? Is each node checking a cache of wallets it has verified and then polling as needed?

- The PeerId previously was providing replay and PITM resistance as the authentication was performed per channel which was tied to a client EC Pair. Do we have a mechanism to tie this token to a particular channel? or minimally a client application? 

- I'm assuming were using TLS for client->gateway security. Are we pinning certificates?

dev/ts/generate
neekolas: You can see the changes I made [here](https://github.com/xmtp/proto/pull/7/commits/4a3ff12b35442c2de0b89ad19bd2cb6a30a69f0e)

Basically, I added the generator to the Docker image and run it as a second codegen step, since the output formats are different. 

This build is getting complicated enough I am starting to wonder if using `buf` would make this more maintainable/readable. I hesitated to include it originally, since it's giving us very little value in `xmtp-js`. But maybe it makes sense here just to make this configuration more declarative.

proto/authn/v1/authn.proto
neekolas: > In this scheme how is the node providing the sybil resistance portion of Authn? Is each node checking a cache of wallets it has verified and then polling as needed?

Think that is the idea. I would imagine the node would look up the wallet transaction history from the Alchemy API on first use of a wallet address, and all rate-limiting would be based on the wallet address and not the token.

> The PeerId previously was providing replay and PITM resistance as the authentication was performed per channel which was tied to a client EC Pair. Do we have a mechanism to tie this token to a particular channel? or minimally a client application?

GRPC Gateway is just a regular HTTP API, we have no other form of client authentication built in. This would be a bearer token. The token would be valid for N minutes, after which it would be rejected. This does mean that a compromised key would be usable by a PITM for N minutes.

mkobetic: > Few Questions:
> 
> * In this scheme how is the node providing the sybil resistance portion of Authn? Is each node checking a cache of wallets it has verified and then polling as needed?

Not quite sure what you mean by sybil resistence here. The limits are tied to a wallet, Authn tied a wallet to a peerId so a request authorization would go peerId -> wallet -> limit. This scheme just skips the peerId bit and ties each requests to a wallet directly. It does that doing the same thing as authn did to tie the peerId to wallet. The downside we do the wallet validation on every request.

> * The PeerId previously was providing replay and PITM resistance as the authentication was performed per channel which was tied to a client EC Pair. Do we have a mechanism to tie this token to a particular channel? or minimally a client application?

There doesn't seem to be a way to tie incoming requests to channels in GRPC world. I'm not sure how to replicate that aspect here. I'm not sure we need to worry about MITM, that should be handled by TLS between the client and the server. The only replay protection with the Token is its expiration period, so if it does get stolen somehow (by a node?) that is an issue for a time. This seems on par with what JWT based authorization does.

We could try to emulate the peerID effect by having the client generate an EC pair, include the public key in the token and sign each request using the private key. That would complicate the message protocol quite a bit. We'd have to repackage each request type into some kind of signed structure. Is that worth it? Do we need to do better than common web auth does?

neekolas: Given that all nodes on the network already have the capability to send unlimited messages, and that this token is only used for sybil resistance/rate limiting of user clients, I'm not concerned about nodes stealing tokens. The horse has already left the barn there.

dev/ts/generate
mkobetic: Thank you! I'd rather postpone switching to buf for a separate PR.

mkobetic:
I updated the PR description to summarize where we are atm (including Jazz's objections). Let me know if something's off.

I'd like to go ahead with this so that I can move forward on the node side. I'd appreciate some ✅ if we are far enough here.

dev/ts/generate
neekolas: Makes sense. I can take that one.

proto/authn/v1/authn.proto
jazzz: @neekolas all sounds reasonable. 
 
 > We could try to emulate the peerID effect by having the client generate an EC pair,
 @mkobetic agreed I don't think this makes sense. I'm thinking forward to the work on malicious client prevention, where we might want a tight coupling of token to application. Non of this should stall this work. 
 
 I think this plan makes sense for the near horizon. As mechanisms such as postage become reality we can migrate to a more privacy preserving method to keeping the network secure.

proto/message_api/v1/authn.proto
```diff
@@ -0,0 +1,66 @@
...
+// AuthData carries token parameters that are authenticated
+// by the identity key signature.
+// It is embedded in the Token structure as bytes
+// so that the bytes don't need to be reconstructed
+// to verify the token signature.
+message AuthData {
+    // address of the wallet
+    string wallet_addr = 1;
+    // time when the token was generated/signed 
+    uint64 timestamp = 2;
```
mkobetic, line 64: One thought: Should we turn this into `valid_until` to allow the client to control the validity expiration within node enforced constraint? (e.g. let's say the nodes will only accept up to an hour or something).

neekolas (approved):

proto/authn/v1/authn.proto
jazzz: > You could generate them to the year 3000 and then use them at the right moment.

If we could include a challenge response, that would be helpful for mitigating future malicious clients . Not sure how feasible that is.

The server provides an input which must be included in the token. In an optimal world it would be unique to a session but I get this this would impact the portability across nodes.

Even if the challenge was a slowly changing network variable (Think block height), old tokens could be invalidated. There's still a window of compromise, but would likely remove the stockpiling issue as it would be less appealing to be storing T*|C| entries

proto/message_api/v1/authn.proto
neekolas: It's hard for me to imagine a client developer choosing anything less than the maximum amount of time, given that the tokens are there to protect XMTP more than they are to protect the user.

mkobetic: I look at it differently. A stolen token is not an issue for the network, it doesn't matter who used up the limit. It is an issue for the user, since it allows someone to DOS them. So I was thinking a more paranoid user may want to reduce the expiration period at the cost of having to create tokens more often.

proto/authn/v1/authn.proto
jazzz: effectively establishing an upper and lower bound on the validity of tokens. This would mean when the rest of the malicious client mitigation work is completed we won't need to worry about the risk of legacy tokens.

mkobetic: Isn't the timestamp basically a "slowly changing network variable"? I don't see stockpiling as an issue. There's no point for a valid client to stockpile tokens. There's nothing to gain from it other then increasing a chance of someone stealing them.

mkobetic: > effectively establishing an upper and lower bound on the validity of tokens. This would mean when the rest of the malicious client mitigation work is completed we won't need to worry about the risk of legacy tokens.

The bounds I want to set with this design are
1. node_time < token.timestamp
2. token.timestamp < node_time + expiration(1h?)

So a token is only valid between token.timestamp and token.timestamp+expiration interpreted in node time.

jazzz: > There's no point for a valid client to stockpile tokens.

Agreed but I'm thinking in the context of a malicious node.

> Isn't the timestamp basically a "slowly changing network variable"?

Time is deterministic, and since the timestamp field is client supplied and untrustworthy theres no guarantee of when that token was generated. By adding a random variable to be included, we build confidence that the token was generated after the challenge was made public.

jazzz: There are other more substantial issues that need to be addressed before this is the limiting factor, so I'm happy to revisit it later. However having a "proof of life" style mechanic that establishes a lowerbound on when the token was generated would reduce risk of token abuse

proto/message_api/v1/authn.proto
neekolas: I still don't see it. If we have a problem where people are able to compromise these tokens and use them for DOS, pretty sure the response is going to be to not use the network at all. Being DOS'd for 20 minutes and 60 feels about the same to the end user: unacceptable.

mkobetic: Fair enough, I'll leave this be.

proto/authn/v1/authn.proto
neekolas: > Agreed but I'm thinking in the context of a malicious client.

At least right now, the malicious client would also have the `IdentityKey`s needed to generate new tokens on demand.

proto/message_api/v1/authn.proto
jazzz: I think having a time the token was generated would be helpful into the future.

Though I could see having an addition field for expiry like in the JWT spec.

neekolas: If we care about having the client be aware of when the tokens are due to expire we could always add a response header where the server would provide that value. Not sure how useful it is given that the client can always just regenerate when a request fails with a 401 status code, but that feels preferable to having the client guess when the token is going to expire.

neekolas:
Great. This is going to be quite easy to integrate into my branch.

mkobetic:
Sigh, as soon as I merged I realized that `AuthData.timestamp` should probably be `AuthData.created_ns` to fit the `message/v1` theme. Worth fixing in a follow up?

I also note that there's also `PublicKey.timestamp`, but changing that would move us away from the xmtp-js stuff. Not quite sure that actually matters though.

mkobetic:
And I think I tripped up with the semantic releaser thingy 🤦

PR #8, opened 2022-08-10
fix: authn cleanup

mkobetic:
Follow up for #7:
* applied proper commit message
* renamed AuthData.timestamp => AuthData.created_ns
* ~~renamed PublicKey.timestamp => PublicKey.created_ns~~
* ~~moved AuthData into Token~~
* moved Secp256k1Uncompresed into PublicKey
* ordered type definitions to have the contents first and inner types after


proto/message_api/v1/authn.proto
```diff
@@ -50,17 +51,19 @@ message Token {
...
-    // address of the wallet
-    string wallet_addr = 1;
-    // time when the token was generated/signed 
-    uint64 timestamp = 2; 
+    // AuthData carries token parameters that are authenticated
+    // by the identity key signature.
+    // It is embedded in the Token structure as bytes
+    // so that the bytes don't need to be reconstructed
+    // to verify the token signature.
+    message AuthData {
```
neekolas, line 60: Why do we want this inside the Token message when it isn't actually used in the Token message?

I'm fine with it being a separate message

proto/message_api/v1/authn.proto
```diff
@@ -13,30 +13,31 @@ option go_package = "./message_api/v1";
...
 message PublicKey {
-    uint64 timestamp = 1;
+    uint64 created_ns = 1;
     Signature signature = 2;  // signature is optional in the xmtp-js version
     oneof union {
         Secp256k1Uncompresed secp256k1_uncompressed = 3;
     }
+
+    // Secp256k1 public key in uncompressed point format.
+    message Secp256k1Uncompresed {
```
neekolas, line 37: This makes sense, given it's used in the message

neekolas:
@mkobetic have you run `npm i` inside the repo? Curious why commitlint actually let you commit without valid messages

proto/message_api/v1/authn.proto
```diff
@@ -13,30 +13,31 @@ option go_package = "./message_api/v1";
...
-message Secp256k1Uncompresed {
-    // uncompressed point with prefix (0x04) [ P || X || Y ], 65 bytes
-    bytes bytes = 1; 
 }
 
 // PublicKey represents a generalized public key,
 // defined as a union to support cryptographic algorithm agility.
 message PublicKey {
-    uint64 timestamp = 1;
+    uint64 created_ns = 1;
```
neekolas, line 30: Given that we are converting back and forth between the existing crypto protos and these, would be nice to have them be as similar as possible.

mkobetic: OK, I can roll this back.

neekolas: Right now the only change is the optional signature, which doesn't prevent the translation between our native crypto types and these new protos directly.

proto/message_api/v1/authn.proto
mkobetic: I was just trying to organize things more uniformly, types that are only used in a message to be inner types of the message. But I can see the argument that AuthData is used indirectly, I can move it back out.

mkobetic:
> @mkobetic have you run `npm i` inside the repo? Curious why commitlint actually let you commit without valid messages

It doesn't let me now, but it did with the previous PR, I'm not sure what was going on there. It may be that I ran `npm i` in the meantime somehow.

neekolas (approved):

proto/message_api/v1/authn.proto
mkobetic: I wasn't clear which proto types will be used for what. But I guess with TS ducktyping they are interchangeable as long as they are structurally the same.

PR #9, opened 2022-08-11
Import client protos

neekolas:
## Summary
This brings in the full set of client protobufs to the `proto` package. There are no changes to the wire format for anything here, although some fields have been renamed to meet our linting rules (mostly around snake casing field names). 

### Contents
- Uses client types for `Signature` and `PublicKey` in Authn
- Namespaces `message_api`
- Exports all client types
- Use `source_relative` go imports so that we can import one package from another
- Added some comments to messages to satisfy the linter

## Notes
- I had to ignore the protolint rules for the `Compression` enum, which does not conform to the rules (it doesn't have an `UNSPECIFIED` 0 value)


mkobetic (approved):
Looks good. 👍

PR #13, opened 2022-08-24
Split up xmtp_envelope.proto

mkobetic:
Given that the public key stuff plays a role on its own in the contact topics and that the xmtp_envelope file is getting long and convoluted, I thought it might be a good idea to break it up in two.

This might require some updates in xmtp-js as some of the types will move into `publicKey.*` in the `index.js`. Other than that it should be pretty harmless. I ran the code generation steps for both go and ts, which seemed to have succeeded.


neekolas:
Given that this is a breaking change for importers of the library, I think we should include a semver commit comment that notes that and causes a major release. We got bit by this a couple weeks ago.

Side note: I previously pinned `@xmtp/proto` to the exact version in `xmtp-js`, but somehow that got reverted (probably by me).

proto/message_contents/public_key.proto
```diff
@@ -0,0 +1,54 @@
...
+// PublicKeyBundle packages the cryptographic keys associated with a wallet,
+// both senders and recipients are identified by their key bundles.
+message PublicKeyBundle {
+    PublicKey identity_key = 1;
+    PublicKey pre_key = 2;
+}
+
+// The PublicKeyBundle of the user on the network.
+// Used to derive shared secrets for encrypted messaging
+message ContactBundleV1 {
```
neekolas, line 45: Do we actually want to break ContactBundle into it's own file too while we are at it?

mkobetic: I'm game, do you mean just the ContactBundle or the PublicKeyBundle as well? If the latter then we should probably break out PrivateKeyBundle as well for consistency. I think I'd slightly prefer to keep it to just the ContactBundle.

go/generate.go
```diff
@@ -8,4 +8,11 @@ import (
 )
 
 //go:generate go install google.golang.org/grpc/cmd/protoc-gen-go-grpc
-//go:generate protoc --proto_path=../proto -I=. -I=../build/tmp/vendor --go_out=paths=source_relative:. --go-grpc_out=paths=source_relative:. --grpc-gateway_out=generate_unbound_methods=true,paths=source_relative:. ../proto/message_api/v1/message_api.proto ../proto/message_api/v1/authn.proto ../proto/message_contents/xmtp_envelope.proto
+//go:generate -command compile protoc --proto_path=../proto -I=. -I=../build/tmp/vendor --go_out=paths=source_relative:. --go-grpc_out=paths=source_relative:. --grpc-gateway_out=generate_unbound_methods=true,paths=source_relative:.
```
mkobetic, line 11: Trying to split up this unmaintainable line (apparently we were already missing private key and composite here). This line defines an "alias" called "compile" that translates to the rest of this line (i.e. all the protoc with all the parameters). Then we can use the alias to compile the proto files individually, making the list of files clearer.

neekolas: We could also have it just run a script that is multi-line

proto/message_contents/public_key.proto
```diff
@@ -0,0 +1,41 @@
+// Public Key
+syntax = "proto3";
+
+package xmtp.message_contents;
+
+option go_package = "github.com/xmtp/proto/go/message_contents";
+
+// Signature represents a generalized public key signature,
+// defined as a union to support cryptographic algorithm agility.
+message Signature {
```
jazzz, line 10: In the future this signature may be used outside of the publicKey message, to sign other objects. I don't think its really worth splitting it out, to figured I would point it out

jazzz (approved):

neekolas (approved):

proto/message_contents/public_key.proto
mkobetic: Hm, I think this might be worth splitting out as well, especially as the number of signature and key types grows. We should already have a second Signature type for wallet signatures. So unless there are objections, I'll split it out too.

go/generate.go
mkobetic: In that case it would probably make more sense to skip go:generate altogether. I'll leave as is, we can decide to switch in separate PR, for now I just wanted to make this more maintainable.

PR #14, opened 2022-08-26
Add protocol spec

mkobetic:
First draft of a spec for the protocol. The idea is for the document to be a guide to the protobuf specs in this repo and fill in any missing bits to provide a reasonably complete description of the protocol.

In theory one should be able to write an SDK that would be compatible with the networks, using just the information in this repository.

I'm purposely omitting private keys and bundles and the private storage topic from the spec as I don't view those as part of the protocol.

I would really like to rename `proto/message_contents/xmtp-envelope.proto` to `proto/message_contents/message.proto` to fit better with the other files there. It is double misleading, because the API `Envelope` is actually in `proto/message_api/v1/message_api.proto`. I would also like to split the file into `message.proto` and `content.proto`. 👍 / 👎  ?


PROTOCOL.md
```diff
@@ -0,0 +1,51 @@
...
+
+### Contacts
+
+To start a conversation the initiator needs the public keys of the participants. To facilitate this step registered clients publish their public key bundle in easily discoverable contact topics.
+
+Each user is automatically assigned a contact topic named `contact-<user's wallet address>`. Public key bundles are published as [ContactBundles](https://github.com/xmtp/proto/blob/main/proto/message_contents/contact.proto) which allow for versioning of the bundle contents and format. A public key bundle MUST be republished whenever the pre-key or identity-key changes. The latest published bundle represents the current bundle that should be used for all new messages and conversations.
+
+(TODO) When publishing a contact into a contact topic, the nodes MUST verify that the bundle being published is a valid bundle signed by the wallet that owns the contact topic (i.e. the address in the topic name matches the address of the wallet that signed the bundle). The nodes must also verify that the created_ns time of the pre-key in the bundle is newer than the created_ns time of the pre-key in the latest published bundle.
```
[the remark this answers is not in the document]
mkobetic: Hm, the TODO marker is a bit misleading, but I'm looking at the spec as a description of what we think the protocol should be, not necessarily a reflection of the current state of our SDK or node software. Of course we should strive to make our SDK and node compliant with the spec and I assume we will, but to me there is a distinction if that makes sense.

proto/message_contents/private_key.proto
```diff
@@ -11,17 +11,16 @@ import "message_contents/public_key.proto";
 
 option go_package = "github.com/xmtp/proto/go/message_contents";
 
-
 // PrivateKey generalized to support different key types
-message PrivateKey {
+message SignedPrivateKey {
```
mkobetic, line 15: The diff here is BS, I simply swapped the legacy types and the V2 types, to move the latter to the top. No actual changes.

PROTOCOL.md
```diff
@@ -0,0 +1,55 @@
+# The XMTP Protocol
+
+## Intro
+
+The protocol allows `clients`, representing identities associated with blockchain accounts (further 'accounts'), to communicate securely with each other. The communications are represented as `conversations` between two or more clients. Conversation consists of a sequence of `messages` from a `sender` to the other `participants` of the conversation, the `recipients`.
```
jhaaaa, line 5:
```suggestion
The protocol allows `clients`, representing identities associated with blockchain accounts (further 'accounts'), to communicate securely with each other. The communications are represented as `conversations` between two or more clients. A conversation consists of a sequence of `messages` from a `sender` to the other `participants` of the conversation, the `recipients`.
```

PROTOCOL.md
```diff
@@ -0,0 +1,55 @@
...
+
+The communication is mediated by a network of `nodes`. The network is responsible for ensuring that messages are delivered between clients. Since clients are not necessarily connected to the network at the same time, the network must be able to store messages while they are in transit.
+
+The protocol is layered to satisfy various operational and security requirements. For example, the clients are responsible for all the encryption and authentication, and key management. The nodes generally see only encrypted content with no indication of identities involved.
+
+## Network Layer
+
+Messages are opaque to the node (usually encrypted), i.e. nodes do not read/interpret the contents of the messages. Unencrypted message headers are readable and can be processed by the nodes.
+
+Consequently messages are just opaque `payload` to the nodes, wrapped in `envelopes` before being passed to the network by the clients. Besides the payload, envelopes also carry a `timestamp` and `topic`.
```
jhaaaa, line 15:
```suggestion
Consequently a message is just an opaque `payload` to the nodes, wrapped in an `envelope` before being passed to the network by the clients. Besides the payload, an envelope also carries a `timestamp` and `topic`.
```

PROTOCOL.md
```diff
@@ -0,0 +1,55 @@
...
+
+Envelopes are always associated with a `topic`. Topics are used to group related envelopes. Topic is a required parameter of all interactions with the network layer, e.g. "add this envelope to this topic", "give me all envelopes from this topic with timestamp between X and Y", or "send me all new envelopes for this topic".
```
jhaaaa, line 17:
```suggestion
Envelopes are always associated with a `topic`. Topics are used to group related envelopes. A topic is a required parameter of all interactions with the network layer, e.g. "add this envelope to this topic", "give me all envelopes from this topic with timestamp between X and Y", or "send me all new envelopes for this topic".
```

PROTOCOL.md
```diff
@@ -0,0 +1,55 @@
...
+
+Nodes employ [Waku relay protocol](https://rfc.vac.dev/spec/11/) to propagate envelopes through the network and [Waku store protocol](https://rfc.vac.dev/spec/13/) to make envelopes persist.
```
jhaaaa, line 19:
```suggestion
Nodes employ the [Waku relay protocol](https://rfc.vac.dev/spec/11/) to propagate envelopes through the network and the [Waku store protocol](https://rfc.vac.dev/spec/13/) to make envelopes persist.
```

PROTOCOL.md
```diff
@@ -0,0 +1,55 @@
...
+
+## Client Layer
+
+Clients connect to an arbitrary node though a GRPC based [message API](https://github.com/xmtp/proto/blob/main/proto/message_api/v1/message_api.proto) and interact with the network as a whole through this node. The message API is the interface separating the client layer from the network layer. Conversations are represented at the network layer as topics.
+
+The client layer uses protobuf for encoding of all its structures. The definitions are maintained in the [xtmp/proto repository](https://github.com/xmtp/proto/blob/main/proto).
```
jhaaaa, line 25:
```suggestion
The client layer uses protobuf for encoding of all its structures. The definitions are maintained in the [xmtp/proto repository](https://github.com/xmtp/proto/blob/main/proto).
```

PROTOCOL.md
```diff
@@ -0,0 +1,55 @@
...
+
+Consequently messages are just opaque `payload` to the nodes, wrapped in `envelopes` before being passed to the network by the clients. Besides the payload, envelopes also carry a `timestamp` and `topic`.
+
+Envelopes are always associated with a `topic`. Topics are used to group related envelopes. Topic is a required parameter of all interactions with the network layer, e.g. "add this envelope to this topic", "give me all envelopes from this topic with timestamp between X and Y", or "send me all new envelopes for this topic".
+
+Nodes employ [Waku relay protocol](https://rfc.vac.dev/spec/11/) to propagate envelopes through the network and [Waku store protocol](https://rfc.vac.dev/spec/13/) to make envelopes persist.
+
+## Client Layer
+
+Clients connect to an arbitrary node though a GRPC based [message API](https://github.com/xmtp/proto/blob/main/proto/message_api/v1/message_api.proto) and interact with the network as a whole through this node. The message API is the interface separating the client layer from the network layer. Conversations are represented at the network layer as topics.
```
jhaaaa, line 23:
```suggestion
Clients connect to an arbitrary node through a GRPC-based [message API](https://github.com/xmtp/proto/blob/main/proto/message_api/v1/message_api.proto) and interact with the network as a whole through this node. The message API is the interface separating the client layer from the network layer. Conversations are represented at the network layer as topics.
```

PROTOCOL.md
```diff
@@ -0,0 +1,55 @@
...
+
+The client layer uses protobuf for encoding of all its structures. The definitions are maintained in the [xtmp/proto repository](https://github.com/xmtp/proto/blob/main/proto).
+
+### Keys
+
+A client is associated with a blockchain account through a set of public/private keys (EC secp256k1) generated when the account is registered with the XMTP network. First there is the identity key that serves as a proxy for the account key and is signed by the account key to establish its authenticity. Then there is one or more pre-keys that are used for message encryption. The pre-keys are signed by the identity key to prove authenticity. The two level key structure is there to facilitate more frequent key rotation without requiring frequent account key signing. The key structure and naming is modelled after [Signal's X3DH protocol](https://signal.org/docs/specifications/x3dh/#the-x3dh-protocol).
```
jhaaaa, line 29:
```suggestion
A client is associated with a blockchain account through a set of public/private keys (EC secp256k1) generated when the account is registered with the XMTP network. First there is the identity key that serves as a proxy for the account key and is signed by the account key to establish its authenticity. Then there are one or more pre-keys that are used for message encryption. The pre-keys are signed by the identity key to prove authenticity. The two-level key structure is there to facilitate more frequent key rotation without requiring frequent account key signing. The key structure and naming are modeled after [Signal's X3DH protocol](https://signal.org/docs/specifications/x3dh/#the-x3dh-protocol).
```

PROTOCOL.md
```diff
@@ -0,0 +1,75 @@
...
+
+The client layer uses protobuf for encoding of all its structures. The definitions are maintained in the [xmtp/proto repository](https://github.com/xmtp/proto/blob/main/proto).
+
+### Keys
+
+A client is associated with a blockchain account through a set of public/private keys (EC secp256k1) generated when the account is registered with the XMTP network. First there is the identity key that serves as a proxy for the account key and is signed by the account key to establish its authenticity. Then there are one or more pre-keys that are used for message encryption. The pre-keys are signed by the identity key to prove authenticity. The two-level key structure is there to facilitate more frequent key rotation without requiring frequent account key signing. The key structure and naming are modeled after [Signal's X3DH protocol](https://signal.org/docs/specifications/x3dh/#the-x3dh-protocol).
```
jhaaaa, line 29: What do you think about this change, @mkobetic? =)

```suggestion
A client is associated with a blockchain account through a set of public/private keys (EC secp256k1) generated when the account is registered with the XMTP network. First there is the identity key that serves as a proxy for the account key and is signed by the account key to establish its authenticity. Then there are one or more pre-keys that are used for invitation encryption. The pre-keys are signed by the identity key to prove authenticity. The two-level key structure is there to facilitate more frequent key rotation without requiring frequent account key signing. The key structure and naming are modeled after [Signal's X3DH protocol](https://signal.org/docs/specifications/x3dh/#the-x3dh-protocol).
```

PROTOCOL.md
```diff
@@ -0,0 +1,75 @@
...
+
+The keys are maintained and stored together in a key bundle, the private keys in a [EncryptedPrivateKeyBundle](https://github.com/xmtp/proto/blob/main/proto/message_contents/private_key.proto) and public keys in a [SignedPublicKeyBundle](https://github.com/xmtp/proto/blob/main/proto/message_contents/public_key.proto).
+
+### Contacts
+
+To start a conversation the initiator needs the public keys of the participants. To facilitate this step registered clients publish their public key bundle in easily discoverable contact topics.
```
jhaaaa, line 35: Super nits!

```suggestion
To start a conversation, the initiator needs the public keys of the participants. To facilitate this step, registered clients publish their public key bundle in easily discoverable contact topics.
```

PROTOCOL.md
```diff
@@ -0,0 +1,75 @@
...
+
+The bytes of the `EncodedContent` are wrapped in [SignedContent](https://github.com/xmtp/proto/blob/main/proto/message_contents/xmtp_envelope.proto), signed using the pre-key of the sender. The signature covers the bytes of the `EncodedContent` and the bytes of the [MessageHeaderV2](https://github.com/xmtp/proto/blob/main/proto/message_contents/xmtp_envelope.proto) to frustrate replay attacks. `SignedContent` is then encrypted using the key material from the invitation and wrapped in [Ciphertext](https://github.com/xmtp/proto/blob/main/proto/message_contents/ciphertext.proto). Finally [MessageV2](https://github.com/xmtp/proto/blob/main/proto/message_contents/xmtp_envelope.proto) combines the `Ciphertext` and the bytes of the message header.
+
+## V1
+
+Previous version of the protocol (further V1) had the differences described below.
+
+The networks support both versions of the protocol in order to allow retrieving historic conversations correctly. V1 conversations can still receive new messages, if the parties initiated them before V2 protocol rollout.
```
jhaaaa, line 61:
```suggestion
The XMTP network supports both versions of the protocol to allow for correct retrieval of historical conversations. V1 conversations can still receive new messages if the parties initiated them before the V2 protocol rollout.
```

jhaaaa: Not sure how you feel about this, Martin - but since we mention V2 here - do we need to acknowledge somewhere that all of the content above the V1 section heading describes XMTP V2?

PROTOCOL.md
```diff
@@ -0,0 +1,75 @@
...
+
+### Keys
+
+The protocol used a different format for keys and key bundles, see [PublicKeyBundle](https://github.com/xmtp/proto/blob/main/proto/message_contents/public_key.proto), although semantically there was very little difference. The old format had optional public key signatures, although this was never actually the case, the signatures were effectively required.
```
jhaaaa, line 65:
```suggestion
The protocol used a different format for keys and key bundles, see [PublicKeyBundle](https://github.com/xmtp/proto/blob/main/proto/message_contents/public_key.proto), although semantically there was very little difference. The old format had optional public key signatures, although this was never actually the case because the signatures were effectively required.
```

PROTOCOL.md
```diff
@@ -0,0 +1,75 @@
...
+
+### Introductions
+
+The protocol didn't use Invitations, instead it used introduction topics named `intro-<account address>`. When a sender wanted to initiate a conversation with a recipient, the client would send the first message to 3 different topics: the sender's intro topic, the recipient's intro topic and the shared conversation topic (see below). This allowed using the intro topics to reconstruct one's list of pre-existing conversations.
```
jhaaaa, line 69: Super nit: 

```suggestion
The protocol didn't use Invitations, instead it used introduction topics named `intro-<account address>`. When a sender wanted to initiate a conversation with a recipient, the client would send the first message to 3 different topics: the sender's intro topic, the recipient's intro topic, and the shared conversation topic (see below). This allowed using the intro topics to reconstruct one's list of pre-existing conversations.
```

PROTOCOL.md
```diff
@@ -0,0 +1,75 @@
...
+
+### Messages
+
+Conversation topics were deterministically named `dm-<account address 1>-<account address 2>` with the two addresses sorted deterministically as well. Consequently there can only be one conversation between two accounts, all messages between these two accounts have to go to this one topic. This structure also allows only one to one conversations. V1 conversations do not have IDs or metadata.
```
jhaaaa, line 73:
```suggestion
Conversation topics were deterministically named `dm-<account address 1>-<account address 2>` with the two addresses sorted deterministically as well. Consequently, there could only be one conversation between two accounts, with all messages between these two accounts stored in a single topic. This structure also allows only one-to-one conversations. V1 conversations do not have IDs or metadata.
```

PROTOCOL.md
```diff
@@ -0,0 +1,75 @@
...
+
+The messages were encrypted using a key derived from the pre-keys of the sender and recipient, a scheme that is identical to the one used for Invitations above. Consequently the sender and the recipient of a message are present in the header, outside of the encrypted payload, and visible to outside observers (i.e. nodes).
```
jhaaaa, line 75:
```suggestion
The messages were encrypted using a key derived from the pre-keys of the sender and recipient, a scheme that is identical to the one used for Invitations above. Consequently, the sender and the recipient of a message were present in the header, outside of the encrypted payload, and visible to outside observers (i.e. nodes).
```

PROTOCOL.md
```diff
@@ -0,0 +1,75 @@
...
+
+### Invitations
+
+A conversation is initiated by a client sending an [Invitation](https://github.com/xmtp/proto/blob/main/proto/message_contents/invitation.proto) to one or more participants. The invitation contains a randomly generated name of the topic to be used for the conversation messages, and randomly generated secret key material to be used for message encryption. It also indicates the encryption algorithm to use for message encryption.
+
+Invitations are published into invite topics. Each blockchain account is automatically assigned an invite topic named `invite-<account address>`. It contains all the invitations both sent and received by the account owner. It allows the client to reconstruct the list of all past conversations and to find their corresponding message topics.
+
+Invitations are encrypted using a key derived from the pre-keys of the invitation sender and recipient. Consequently the sender and the recipient of an invitation are present in the header, outside of the encrypted payload, and visible to outside observers (i.e. nodes).
```
jhaaaa, line 47:
```suggestion
Invitations are encrypted using a key derived from the pre-keys of the invitation sender and recipient. Consequently, the sender and the recipient of an invitation are present in the header, outside of the encrypted payload, and visible to outside observers (i.e. nodes).
```

PROTOCOL.md
```diff
@@ -0,0 +1,75 @@
...
+
+The communication is mediated by a network of `nodes`. The network is responsible for ensuring that messages are delivered between clients. Since clients are not necessarily connected to the network at the same time, the network must be able to store messages while they are in transit.
+
+The protocol is layered to satisfy various operational and security requirements. For example, the clients are responsible for all the encryption and authentication, and key management. The nodes generally see only encrypted content with no indication of identities involved.
+
+## Network Layer
+
+Messages are opaque to the node (usually encrypted), i.e. nodes do not read/interpret the contents of the messages. Unencrypted message headers are readable and can be processed by the nodes.
+
+Consequently a message is just an opaque `payload` to the nodes, wrapped in an `envelope` before being passed to the network by the clients. Besides the payload, an envelope also carries a `timestamp` and `topic`.
```
jhaaaa, line 15:
```suggestion
Consequently, a message is just an opaque `payload` to the nodes, wrapped in an `envelope` before being passed to the network by the clients. Besides the payload, an envelope also carries a `timestamp` and `topic`.
```

PROTOCOL.md
mkobetic: Hm, you are right @jhaaaa that could be misleading. But I want to be a bit more general/vague here. So how about I change `message encryption` to `payload encryption and signing`?

PROTOCOL.md
mkobetic: I agree, we should. But I'm struggling with how. Ideally it would be right at the top of the document, but it doesn't fit well with the way the Intro is written.  Do you think it could go above the Intro, just say something like.

> Current version of the protocol is V2. This is what this document focuses on. Prior versions are described at the end of the document.

PROTOCOL.md
```diff
@@ -0,0 +1,75 @@
+# The XMTP Protocol
+
+## Intro
+
+The protocol allows `clients`, representing identities associated with blockchain accounts (further 'accounts'), to communicate securely with each other. The communications are represented as `conversations` between two or more clients. A conversation consists of a sequence of `messages` from a `sender` to the other `participants` of the conversation, the `recipients`.
+
+The communication is mediated by a network of `nodes`. The network is responsible for ensuring that messages are delivered between clients. Since clients are not necessarily connected to the network at the same time, the network must be able to store messages while they are in transit.
+
+The protocol is layered to satisfy various operational and security requirements. For example, the clients are responsible for all the encryption and authentication, and key management. The nodes generally see only encrypted content with no indication of identities involved.
```
elisealix22, line 9: The word "generally" here makes me think the nodes might see unencrypted data. I know you explain it below, but I'm wondering if we can make this last sentence flex a little harder.

Maybe something like `The nodes see only encrypted user-generated content with no indication of identities involved`?

PROTOCOL.md
```diff
@@ -0,0 +1,75 @@
...
+
+## Network Layer
+
+Messages are opaque to the node (usually encrypted), i.e. nodes do not read/interpret the contents of the messages. Unencrypted message headers are readable and can be processed by the nodes.
```
elisealix22, line 13: Same here with "usually" -- when are messages unencrypted? Seems like it still works without it:

Messages are opaque to the node (encrypted), i.e. nodes do not read/interpret the contents of the messages. Unencrypted message headers are readable and can be processed by the nodes.

elisealix22 (commented):
Martin, you have no idea how much this helped ME understand our protocol. Seriously, amazing work. My only note is if we can remove the words "generally" and "usually", then I think it will be much more clear and leave less room for follow-up questions.

PROTOCOL.md
mkobetic: I like that.

PROTOCOL.md
mkobetic: I'm hedging because few paragraphs below I'm proposing the nodes to actually parse and validate contact bundles. Maybe that's not really necessary. Maybe I should drop that for now, @neekolas was pushing against that as well. But even then I'm a bit iffy about a categoric statement that nodes will not interpret the payload under any circumstances.

mkobetic: Then again, given the terminology I'm pushing here where `message` means payload with user generated content, then those would always be encrypted. I'll review and make sure to use `payload` for arbitrary envelope content and `message` for payload with user generated content.

mkobetic:
I want to highlight this question from the description, any yays or nays?

> I would really like to rename `proto/message_contents/xmtp-envelope.proto` to `proto/message_contents/message.proto` to fit better with the other files there. It is double misleading, because the API `Envelope` is actually in `proto/message_api/v1/message_api.proto`. I would also like to split the file into `message.proto` and `content.proto`. 👍 / 👎 ?

PROTOCOL.md
elisealix22: >I'm hedging because few paragraphs below I'm proposing the nodes to actually parse and validate contact bundles.

I think the piece that might be confusing is does that go against the next part of the sentence, "i.e. nodes do not read/interpret the contents of the messages"? Maybe we can untangle this by listing out what is encrypted and what is not -- almost like a series of facts instead of a general statement? Idk if that introduces too much complexity early on though and I def trust your gut here! I don't want to pigeon-hole us.

PROTOCOL.md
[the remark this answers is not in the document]
mkobetic: I think I'm going to drop this bit after all. It can be brought back as a future revision if we end up agreeing that this is a good idea.

elisealix22 (approved):
This is an incredible milestone in my book 🥲

PROTOCOL.md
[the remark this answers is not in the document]
mkobetic: I replaced the paragraph with

> Only account owners SHOULD be able to publish into their contact topics. This could be enforced at the API or at the node level. Specifics are out of scope in this document.

We can revise later as we see fit.

mkobetic:
I will leave the proposed file rearrangements for a follow up PR to avoid delaying this PR.

PROTOCOL.md
jhaaaa: Thank you for your patient consideration, @mkobetic! I like your suggested change!

PROTOCOL.md
jhaaaa: I think right above the "Intro" is a great place for this information, @mkobetic! Perhaps exactly what you wrote, with just a few tweaks - something like this?

>This document focuses on describing the current version of the Extensible Message Transport Protocol: XMTP V2.
>
>To learn about the previous version of the protocol, see [V1](#V1).

mkobetic: Love it!

mkobetic:
Last bit I'd like to include here is updating README.md with some sort of blurb that will point to the new PROTOCOL.md file. I think I'm hitting a writers block, so if anyone has suggestions, I'd welcome some pointers.

mkobetic:
OK, I've got something in the README, it's not great but would probably do. Welcome any feedback/suggestions.

PROTOCOL.md
```diff
@@ -0,0 +1,81 @@
...
+
+A public key bundle MUST be republished whenever the pre-key or identity-key changes. The latest published bundle represents the current bundle that should be used for all new messages and conversations.
+
+Only account owners SHOULD be able to publish into their contact topics. This could be enforced at the API or at the node level. Specifics are out of scope in this document.
+
+### Invitations
+
+A conversation is initiated by a client sending an [Invitation](https://github.com/xmtp/proto/blob/main/proto/message_contents/invitation.proto) to one or more participants. The invitation contains a randomly generated name of the topic to be used for the conversation messages, and randomly generated secret key material to be used for message encryption. It also indicates the encryption algorithm to use for message encryption.
```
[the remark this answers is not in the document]
mkobetic: > Two questions related to the conversation topic name
> 1. How do we avoid name squatting?

Not quite sure what you mean here.

> 2. What happens during a collision?

Unintentional collision resistance is achieved by prescribing that the topic MUST be randomly generated 32 bytes https://github.com/xmtp/proto/blob/main/proto/message_contents/invitation.proto#L29-L31.

There is nothing preventing intentional collision. If someone issues an invitation with the same topic, we'll have two separate conversations mixed together, presumably the different parties unable to read other party messagese assuming different keys are used.

PR #15, opened 2022-08-30
Negotiated Topics

### LINKED ISSUE #10: PublicKey vs SignedPublicKey
PublicKey signature is I think the only case where we don't capture the signed bytes. Consequently signature verification requires remarshaling the key contents (without the signature), which we know can be fraught with non-determinism in protobuf. We may want to consider replacing the current PublicKey definition
```
message PublicKey {
    // The key bytes
    message Secp256k1Uncompressed {
        // uncompressed point with prefix (0x04) [ P || X || Y ], 65 bytes
        bytes bytes = 1; 
    }
    uint64 timestamp = 1;
    optional Signature signature = 2;
    oneof union {
        Secp256k1Uncompressed secp256k1_uncompressed = 3;
    }
}
```
With something like this
```
message PublicKey {
    // The key bytes
    message Secp256k1Uncompressed {
        // uncompressed point with prefix (0x04) [ P || X || Y ], 65 bytes
        bytes bytes = 1; 
    }
    uint64 timestamp = 1;
    oneof union {
        Secp256k1Uncompressed secp256k1_uncompressed = 3;
    }
}

message SignedPublicKey {
    bytes key_bytes = 1;
    Signature signature = 2;
}
```

Upside:
* safe from protobuf issues
* removes the optional field which seems to be a hassle
* saves remarshaling for the verifier

Downside:
* this would be a breaking change, all the contact bundles would need to be re-exported

Thoughts?

mkobetic:
Implements [Negotiated Topics RFC](https://github.com/xmtp-labs/hq/pull/661) and relevant bits of [Key Passing RFC](https://github.com/xmtp-labs/hq/pull/662). Also fixes #10.

This will require extensive changes in xmtp-js, largely to handle the multiple versions of things.

Will squash/rebase as a BREAKING CHANGE.


proto/message_contents/ciphertext.proto
```diff
@@ -0,0 +1,20 @@
+// Ciphertext
```
mkobetic, line 1: Extracted Ciphertext as well, as it is used in multiple places and it seems wrong to import xmtp_envelope for it.

proto/message_contents/invitation.proto
```diff
@@ -0,0 +1,43 @@
...
+package xmtp.message_contents;
+
+import "message_contents/ciphertext.proto";
+import "message_contents/public_key.proto";
+
+option go_package = "github.com/xmtp/proto/go/message_contents";
+
+// Unsealed invitation V1
+message InvitationV1 {
+    bytes topic = 1; // randomly generated topic name (>= 32 bytes)
```
mkobetic, line 13: This reflects the proposal from the Key Passing RFC.

proto/message_contents/signature.proto
```diff
@@ -8,12 +8,21 @@ option go_package = "github.com/xmtp/proto/go/message_contents";
...
+    // ECDSA signature bytes and the recovery bit
     message ECDSACompact {
         bytes bytes = 1; // compact representation [ R || S ], 64 bytes
         uint32 recovery = 2; // recovery bit
     }
+    // ECDSA signature bytes and the recovery bit
+    // produced by xmtp-js::PublicKey.signWithWallet function, i.e.
+    // EIP-191 signature of a "Create Identity" message with the key embedded.
+    // Used to sign identity keys.
+    message WalletECDSACompact {
```
mkobetic, line 20: We should start using this for identity key signatures. Maybe we can do that for the new SignedPublicKeys only to make the logic easier to handle.

mkobetic:
Oops, missing the message signing still.

proto/message_contents/invitation.proto
neekolas: Given that the client SDK treats topics as strings everywhere, it would be helpful for this to also be a string. Obviously we can base64 encode it when we use it later, but it seems more straightforward to handle that encoding/decoding once at the time of creation.

proto/message_contents/public_key.proto
```diff
@@ -7,6 +7,34 @@ import "message_contents/signature.proto";
...
+message SignedPublicKey {
+    bytes key_bytes = 1;  // embeds an UnsignedPublicKey
+    Signature signature = 2; // signs key_bytes
+}
+
+// PublicKeyBundle packages the cryptographic keys associated with a wallet,
+// both senders and recipients are identified by their key bundles.
+message SignedPublicKeyBundle {
+    SignedPublicKey identity_key = 1;
+    SignedPublicKey pre_key = 2;
```
neekolas, line 34: The `IdentityKey` is signed with the wallet. Is the `pre_key` expected to be signed with the `IdentityKey`? If so, we should note that with a comment.

proto/message_contents/invitation.proto
mkobetic: OK, so the topic will be generated as 32 random bytes and then base64 encoded (unpadded). Sound good?

What is the easiest thing to do for the message signature? Just take the ascii bytes to concatenate with the header and payload?

proto/message_contents/invitation.proto
```diff
@@ -0,0 +1,44 @@
...
+// Sealed Invitation V1 Header
+message SealedInvitationHeaderV1 {
+    SignedPublicKeyBundle sender = 1;
+    SignedPublicKeyBundle recipient = 2;
+    uint64 created_ns = 3;
+}
+
+// Sealed Invitation V1
+message SealedInvitationV1 {
+    // encoded SealedInvitationHeader used as associated data for the ciphertext
```
neekolas, line 33: I don't see `SealedInvitationHeader` anywhere. I just see `SealedInvitationHeaderV1`

proto/message_contents/invitation.proto
```diff
@@ -0,0 +1,44 @@
...
+    bytes header_bytes = 1;
```
neekolas, line 34: I don't see a field for the header signature. Don't we need that?

proto/message_contents/invitation.proto
neekolas: That sounds reasonable

proto/message_contents/signature.proto
neekolas: Sounds good to me. We are already going to have to refactor the `ContactBundle` for V2

neekolas (approved):
I like these changes

proto/message_contents/xmtp_envelope.proto
```diff
@@ -69,12 +67,30 @@ message MessageHeader {
...
     Ciphertext ciphertext = 2;
 }
 
-// The outermost wrapper for a message in a conversation topic
+
+// Versioned Message
 message Message {
     oneof version {
         V1Message v1 = 1;
+        MessageV2 v2 = 2;
```
neekolas, line 94: A little weird that we have `V1Message` and `MessageV2`. While we are at it do we not just want to rename `V1Message` to `MessageV1` for consistency? It won't change the wire format.

proto/message_contents/xmtp_envelope.proto
```diff
@@ -42,23 +43,20 @@ message EncodedContent {
...
-        Aes256gcmHkdfsha256 aes256_gcm_hkdf_sha256 = 1;
-    }
+// SignedContent
+message SignedContent {
+    // MUST contain EncodedContent
+    bytes payload = 1;
+    SignedPublicKeyBundle sender = 2;
+    // MUST be a signature of a concatenation of the topic name,
+    // message header bytes and the payload bytes,
+    // signed by the sender
```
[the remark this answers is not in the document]
mkobetic: Yup, I'm going over the comments now and beefing them up, will include this too.

mkobetic:
@neekolas, could you take a look at the test failure? Both `npm ci` and `npm test` are passing for me locally. Not sure what this is about. 🙏

mkobetic:
> @neekolas, could you take a look at the test failure? Both `npm ci` and `npm test` are passing for me locally. Not sure what this is about. 🙏

Hm, I suspect it's passing locally because my docker image for xmtp/protoc is cached whereas the CI rebuilds it from scratch. I'm guessing `npm i -g ts-proto` broke recently or something. I'm going to rebase and merge this, but I don't think we can merge this without fixing `npm ci` as it's part of the release step.

neekolas:
Just had to step out for a bit, but I agree we most likely just need to lock the version of ts-proto

ts/index.ts
```diff
@@ -6,4 +6,6 @@ export * as composite from "./message_contents/composite.pb";
 export * as privateKey from "./message_contents/private_key.pb";
 export * as publicKey from "./message_contents/public_key.pb";
 export * as signature from "./message_contents/signature.pb";
+export * as ciphertext from "./message_contents/ciphertext.pb";
 export * as contact from "./message_contents/contact.pb";
+export * as invitation from "./message_contents/invitation.pb";
```
mkobetic, line 11: @neekolas, does this mean that all these types will have to be referenced as `ciphertext.Ciphertext` and `invitation.Invitation`?  Is there a way to export these as a single namespace like we have in Go?

neekolas: You could do `export * from ...`, which would expose the entire namespace. Does open up the possibility of collisions. Still might make sense for the smaller modules.

mkobetic: Hm, maybe not, we're mixing multiple packages here. I guess I'll leave it as is?

neekolas: I think it's fine as is. Will be consistent at least. Everything can be found with $package.$message, as opposed to some message types being exposed at the top level and some being namespaced.

PR #18, opened 2022-09-29
Add context to invite

neekolas:
## Summary

Adds `context` field to invite in line with proposal found here https://github.com/xmtp-labs/hq/issues/690.

Treating this as a non-breaking change, as the protos are not yet in use.


proto/message_contents/invitation.proto
```diff
@@ -12,25 +12,36 @@ option go_package = "github.com/xmtp/proto/go/message_contents";
...
+    // Supported encryption schemes
+    // AES256-GCM-HKDF-SHA256
+    message Aes256gcmHkdfsha256 {
+        bytes key_material = 1; // randomly generated key material (32 bytes)
+    }
+
+    // The Tag type
+    message Context {
+        // Expected to be a URI (ie xmtp.org/convo1)
+        string conversation_key = 1;
```
neekolas, line 24: I've been struggling with the name here. I originally had it as `conversation_id`, but that sends the wrong message that this should be globally unique. It is perfectly valid for a developer to use the same value on every conversation in their app.

Key is an overloaded word at XMTP, but is the most commonly used term in group by expressions (which is really what this is). The purpose of this value is to group together related invites from the same address, and to be queried against in filtering those invites. 

Open to suggestions.

mkobetic: I commented on this in response to your write-up. Looks like we are on the same page wrt to the purpose of the value. I suggested `type`, as a sort of an analog to content type maybe? Not sure that's a good analogy to make though.

proto/message_contents/invitation.proto
```diff
@@ -12,25 +12,36 @@ option go_package = "github.com/xmtp/proto/go/message_contents";
...
 message InvitationV1 {
+    // Supported encryption schemes
+    // AES256-GCM-HKDF-SHA256
+    message Aes256gcmHkdfsha256 {
+        bytes key_material = 1; // randomly generated key material (32 bytes)
+    }
+
+    // The Tag type
+    message Context {
+        // Expected to be a URI (ie xmtp.org/convo1)
```
mkobetic, line 23:
```suggestion
        // Groups related conversations together
        // Can be used to filter user conversations
        // Expected to be a URI (ie xmtp.org/convo1)
```

proto/message_contents/invitation.proto
neekolas: `conversation_type` is probably better than either of my suggestions

neekolas: It does still feel a little odd to describe `bosonprotocol.io/items/item12312312` as a "type" of conversation. But I would imagine things like that would be one of the most common use-cases.

mkobetic: Can you elaborate on the boson use case? I don't quite understand it.

neekolas: So, for Boson they want to group conversations based on a particular item for sale on their platform. They are already using content types to do this in a hacky way. The system (regardless of name) would work great for that using the format above.

mkobetic: Hm, they could also just drop the item id into the metadata. We wouldn't filter it all out for them, but they can easily filter further using metadata. That feels like the more appropriate thing to do in this case. Maybe calling the field `type` would nudge them more towards metadata too.

neekolas: The reason I am pushing for all filtering to be based on a single field is that it simplifies the mapping of invitations -> conversations. All invites with the same user/identifier combo will be grouped together into a conversation, always. When you try and "find or create" a second time it is clear that it will find the existing convo.

Querying based on metadata leaves the grouping ambiguous. What happens when there are multiple keys in the metadata object and only one changes? Does that result in a new conversation or not?

mkobetic: > The reason I am pushing for all filtering to be based on a single field is that it simplifies the mapping of invitations -> conversations. All invites with the same user/identifier combo will be grouped together into a conversation, always. When you try and "find or create" a second time it is clear that it will find the existing convo.

My mental model was that mapping of invitations to conversations will be 1:1 (doesn't get more simple than that). The user can decide to stitch conversations together on their end if they want to, but on our end the picture could be as simple as that. I'm not yet seeing a convincing argument to make it more complex than that.

> Querying based on metadata leaves the grouping ambiguous. What happens when there are multiple keys in the metadata object and only one changes? Does that result in a new conversation or not?

I don't want to provide querying filters based on metadata. I would expect that sort of filtering to happen in user code. It won't be any less efficient than if we did it for them, possibly more.

neekolas: > My mental model was that mapping of invitations to conversations will be 1:1 (doesn't get more simple than that). The user can decide to stitch conversations together on their end if they want to, but on our end the picture could be as simple as that. I'm not yet seeing a convincing argument to make it more complex than that.

How will that work once we have topic rotation? It feels like we are going to be living in a world with multiple invitations per conversation no matter what.

mkobetic: > > My mental model was that mapping of invitations to conversations will be 1:1 (doesn't get more simple than that). The user can decide to stitch conversations together on their end if they want to, but on our end the picture could be as simple as that. I'm not yet seeing a convincing argument to make it more complex than that.
> 
> How will that work once we have topic rotation? It feels like we are going to be living in a world with multiple invitations per conversation no matter what.

By topic rotation you mean starting a new topic so that the key can be rotated? I'm not sure, I'm inclined to leave that to the chat client, rather than trying to provide it at the sdk level. I'm not sure we can come up with something generic enough to fit well in most use cases.

neekolas: > I don't want to provide querying filters based on metadata. I would expect that sort of filtering to happen in user code. It won't be any less efficient than if we did it for them, possibly more.

Neither do I. If someone wants to do some sort of querying based on metadata, they can iterate through all the conversations and DIY it. But I would like for that to be a rare occurrence, not the default case of any time a developer wants to open a conversation with a filter applied to the list.

mkobetic: > Neither do I. If someone wants to do some sort of querying based on metadata, they can iterate through all the conversations and DIY it. But I would like for that to be a rare occurrence, not the default case. That's why I feel strongly we should build some level of invite -> conversation grouping into the SDK

OK, I think I'm starting to see where you were coming from with the `conversation_id`. It makes total sense as a way for stitching multiple invitations/topics into a single conversation. That would be the way to do it if we want that. I think the confusing part is that we are also trying to give it the responsibility for conversation filtering. My first reaction would be to keep the two things separate, i.e. keep the type separate from the convo id, but then you have the possibility of the type being different between multiple invites with the same convo ID. The type really should be associated with the conversation as  a whole and not with the individual invites.

This all makes me worried that the stitched conversation model may end up with all kinds of odd edge cases that we'll have to explain and live with. Thus the appeal of the simpler model without stitching.

I'd like to at least postpone the stitching stuff until we have at least some hands on experience with doing it on the app side. Although I have to admit I'm not sure how hard it's going to be to add that into the SDK later.

mkobetic: I also worry about detrimental effects of stitched conversations on the implementation side. If all conversations are potentially stitched, then we will always have to read all invites in order to reliably reassemble any particular conversation. If we do go this route, we may want to add some sort of a flag on the invite to indicate that it's a part of a larger conversation or something.

mkobetic: Also do all the invites of a conversation be from the same sender? That would mean that only the initiator of a conversation can create new invites, is that what we want?

neekolas: OK. At least for the purposes of this Github Discussions post, I can strip out the stitching stuff and see how people react. We could still have a `findOrCreate` type function, but it would not do any sort of stitching. Just find a topic that matches or create one. Metadata would be totally permanent for a conversation.

We can discuss more next week and decide if we really need stitching or not. 

I'm back to thinking `conversation_id` was an appropriate term and something like that gives us the most optionality going forward. `conversation_type` doesn't feel appropriate for some of the use cases, like the way Lens would use it or the Boson example above. How about `conversation_identifier`, which I know is a synonym, but I think gives less of an impression that it should be globally unique and act like a database ID.

mkobetic: Hm, I don't think `conversation_identifier` is better than `conversation_id`. It also doesn't make much sense if we don't do stitching. In the simple model `type` makes more sense to me.

neekolas: Even in the simple model, I would want developers to include all the data that could be used for filtering into the field. Otherwise we lose the optionality of adding stitching later. 

A bunch of conversation types that are just domains aren't very useful.

mkobetic: Fair enough, I don't have any more arguments, so let's go with your pick and see what the feedback is.

neekolas:
@mkobetic thoughts on merging this as-is so I can prototype an implementation

mkobetic (approved):

PR #19, opened 2022-10-04
feat: add topic to message v2 header

mkobetic:
In https://github.com/xmtp-labs/hq/issues/679 we've agreed on following changes:

1) Add `topic` to `MessageHeaderV2`

2) Change the `SignedContent.signature` definition to be: a signature of a concatenation of message header bytes and the payload bytes

This should reduce harmful impact of message replay when we introduce `SignedContent`


proto/message_api/v1/message_api.proto
```diff
@@ -67,7 +67,11 @@ message PagingInfo {
 
 // Envelope encapsulates a message while in transit.
 message Envelope {
+    // The topic the message belongs to,
+    // MUST be the same as the topic embedded in the message.
     string content_topic = 1;
+    // Message timestamp
+    // MUST be the same as the timestamp embedded in the message.
```
neekolas, line 74: MUST might be too strong here. That's only true for the newer types. We do still have other things that fit into the envelope that do not have conforming timestamps

neekolas: 👍 

I guess the confusing part is that a lot of the old message types do have timestamps. They just are in milliseconds

mkobetic: OK, I can clarify "the same" as "equivalent" maybe? Is that good enough?

neekolas: Works for me

neekolas: I'm also going to fix the issue that causes them to be slightly different for existing messages

mkobetic: > I'm also going to fix the issue that causes them to be slightly different for existing messages

I just started looking into that.

neekolas: You can have that one

PR #21, opened 2022-11-09
feat: generate dart code

dmccartney:
This generates and prepares a dart package for publication.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -31,3 +31,14 @@ jobs:
         with:
           version: v1.45.2
           args: --timeout=5m
+  dart:
+    name: Lint (dart)
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: dart-lang/setup-dart@v1.3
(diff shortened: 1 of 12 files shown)


.husky/pre-commit
```diff
@@ -1,5 +1,6 @@
 #!/usr/bin/env sh
 . "$(dirname -- "$0")/_/husky.sh"
 
-npm test
-./dev/go/generate
+# TODO: cut pre-commit or uncomment before merging
+#npm test
+#./dev/go/generate
```
dmccartney, line 6: the pre-commit hook is really heavy -- I wonder if we need it?

nakajima (commented):
Wouldn't say I'm qualified to review the dart stuff, but otherwise looks good to me!

mkobetic (approved):
I don't know anything about dart, but the overall structure looks good to me. Note that I'm proposing a small refactor in #24 which is likely to impact this, but I can fix that up if we decide to go with it.

elisealix22 (approved):
Looks good to me!

PR #22, opened 2022-11-09
build: cut precommit hook, introduce proto-change test

dmccartney:


.husky/commit-msg
```diff
@@ -1,4 +0,0 @@
-#!/usr/bin/env sh
-. "$(dirname -- "$0")/_/husky.sh"
-
-npx commitlint --edit
```
neekolas, line 4: We still need this to enforce the commit message conventions, which are used by semantic-release.

neekolas (changes requested):
I'm good with removing the build hook, but we still need the commit message hook. Otherwise commits won't generate releases on NPM.

.husky/commit-msg
neekolas: Now that this is going to be generating code for more platforms, we are probably going to have to rethink our conventions for semantic-release (should a breaking change to the Dart build really trigger a new release on the JS package? Probably not). But I think that for the time-being we should leave things as-is and update when we have a reasonable alternative.

neekolas (approved):

dmccartney:
This isn't quite working: the `protoc` for `generate.go` inside the `proto-change-test` workflow is having versioning and path issues (so it's giving us false-negatives -- it's failing when a proto file changes but not because of an unexpected diff). 

Instead of fighting through the go env quirks of this, I'm inclined to close this and revisit when time permits us to refactor codegen more generally. (But if someone else is inspired, feel free to pick this up and run with it.)

PR #23, opened 2022-11-09
feat: generate swift code

nakajima:
Generates swift files. I'm checking them in because that's how SPM seems to want it ([similar to go](https://github.com/xmtp/proto/pull/4#issuecomment-1198249166)).

One small thing I ran into is that I wasn't able to use the `macos` GitHub action environment for tests because that doesn't include Docker ([more here](https://github.com/orgs/community/discussions/25777#discussioncomment-3249229)). While I think most people (everyone?) using this library will be on a Mac, there's nothing really platform specific about the generated code so I think we're ok here.

I was able to import the types into a [sample Xcode project](https://user-images.githubusercontent.com/483/201171196-4fcfe84a-c4e3-4958-a1d4-11e220c9b966.png) from the `swift` branch, once this lands, people will be able to point to `main` but it'd be nice to tag it so people can have something more concrete to point to. I'm not sure how y'all like to handle tagging though so let me know what works best for you.


dev/swift/protoc/Dockerfile
```diff
@@ -0,0 +1,23 @@
+FROM swift:5.7
+
+COPY protoc.sh /opt/protoc
+
+RUN apt-get -q update && \
+    apt-get -y install protobuf-compiler unzip curl \
+    && rm -r /var/lib/apt/lists/*
```
neekolas, line 7: Might want to align the && to either be on the beginning or ends of the lines for readability. Otherwise LGTM

neekolas (approved):

neekolas:
This looks great. Should probably add a similar test to https://github.com/xmtp/proto/pull/22 to ensure the generated files are current.

nakajima:
Fixed up the formatting suggestion. Looks like #22 was closed but I can take a look at getting something swift specific in place in a follow up if that's the approach we want to take!

PR #24, opened 2022-11-10
xmtp_envelope.proto rename/split

mkobetic:
This is the file rename/split I proposed in #14. The rationale is that envelope is really message_api concept, and that the file itself is getting long. The proposal is to split it into a `message.proto` with all the message/header types, and `content.proto` with all the message content types.

This should be transparent for go clients, but ts clients will need to adjust their imports since the `xmtpEnvelope` export is now split into `message` and `content` imports. That's a bit of friction for sure. Worth it?

If we go with this I'll follow up with `xmpt-js` update promptly


neekolas (approved):

mkobetic:
@dmccartney, @nakajima this change may be somewhat disruptive to the swift and dart dependencies. I will update xmtp-js, to change any `xmtpEnvelope` imports from `@xmtp/proto` to be either `message` or `content` as appropriate. Wondering if something like that will be needed for the swift or dart clients.

nakajima:
As long as the type names are the same, things should just work in Swift land I think.

mkobetic:
> As long as the type names are the same, things should just work in Swift land I think.

OK, good. Just wanted to make you aware if things were more complicated, like in xmtp-js.

PR #25, opened 2022-11-14
Add subscribe all endpoint

neekolas:
## Summary

For the Push Server, we are adding a SubscribeAll endpoint that will deliver a firehose of all messages. While it is conceivable to extend the existing Subscribe endpoint, I think it makes sense to make it standalone. The rationale for a new endpoint is:

1. Doesn't muddy up the existing subscribe API with new options, or odd behaviour (sending an empty list of topics gives you the firehose)
2. The implementation will be quite different (no need for a dispatcher or subscriptions, just hook in to the Relay broadcaster)

Resolves https://github.com/xmtp-labs/hq/issues/837


mkobetic (approved):
I don't know whether the swift and dart code should be included, but this is otherwise pretty straightforward.

PR #27, opened 2022-11-15
feat: generate gRPC types for swift

nakajima:
We'll want these to actually use the types generated in https://github.com/xmtp/proto/pull/23.


neekolas (approved):

nakajima:
@neekolas Thanks!

PR #30, opened 2022-11-24
message API: UpdateSubscription

mkobetic:
This API should help fixing the race conditions in the `Conversations.streamAllMessages()` functionality. To implement this on the SDK side, the function has to unsubscribe and resubscribe every time a new conversation topic shows up. This creates holes during which it can miss messages. It can potentially miss even invites, which means any messages for those conversations will be missed as well.

The solution proposed here is to provide an additional api call that can be used to update an existing subscription (a stream in progress). We need to be able to identify the pre-existing subscription, thus the Subscribe api call is extended to allow the client to provide a subscription ID.

Having the client provide the ID is a bit clumsy, but we cannot return a server assigned ID from the streaming call, unless it was carried by the first envelope returned by the stream or something. That would however be a protocol breaking change. Might be something to consider, but this proposal is trying to do it in backward compatible way.

It would be important that the clients generate reasonably random/unpredictable IDs to prevent other clients interfering with it. We could potentially scope the IDs by wallet address, but that would require turning on authentication for Subscribe requests. Worth it?

The server side implementation prototype is here https://github.com/xmtp/xmtp-node-go/pull/191. I'll need these proto changes before I can continue with the prototype.


neekolas:
I like this in concept, and am not super worried about collisions on IDs.

Are we imagining that the subscriptions are going to have some sort of state in the database? Given that our services are always accessed behind a load balancer, there is no guarantee that the request to update the subscription will be sent to the same node that is handling the subscription.

mkobetic:
> Are we imagining that the subscriptions are going to have some sort of state in the database? Given that our services are always accessed behind a load balancer, there is no guarantee that the request to update the subscription will be sent to the same node that is handling the subscription.

No I definitely didn't intend to link nodes through the database. And I have indeed completely forgotten about the loadbalancer issue (again). But I'm wondering if this is more of a short term issue, since eventually clients will likely need to connect to the nodes directly once nodes are not necessarily affiliated and interchangeable. In that world this sort of thing makes more sense I think. Could we maybe turn on node affinity on the LB to make our current setup closer to the future?

neekolas:
We can definitely turn on sticky sessions on the LB. No idea how strong the guarantees are there. Streams can be open for a very long time. 

Worth experimenting with at least.

package.json
```diff
@@ -55,7 +55,8 @@
     "@commitlint/config-conventional": "^17.0.3",
     "@types/node": "^18.6.4",
     "husky": "^8.0.0",
-    "semantic-release": "^19.0.3"
+    "semantic-release": "^19.0.3",
+    "typescript": "^4.6.4"
```
mkobetic, line 59: Without this the CI was failing with a mismatch between node and dom setup around AbortSignal. This mismatch seems to be introduced with [TS 4.9](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/62782). Which is what the CI probably upgraded to. Locally with TS 4.6 I couldn't reproduce the problem.

elisealix22 (approved):
I LOVE this direction and think @dmccartney and @nakajima will likely want this API when implementing `Conversations.streamAllMessages()` as well!

mkobetic:
Another thought re the LB issues, HTTP2 persistent connections could probably work around a lot of that too.

neekolas:
@mkobetic Would have to confirm support in GRPC gateway and browser Fetch API

dmccartney:
Quick thoughts:
- I love that we're gonna hammer out the expected client subscription flows.
- I'm trying to think through when a real app would ever `SubscribeAll` for multiple conversations in a single shot. I think most subscriptions would be in the context of a single or small number of conversations. The streaming seems most useful to enable a real-time-ish UX and reduce `Query` calls. But I think we should still rely on `Query` to ensure the local store of the conversation is eventually consistent with the global state.

E.g. I'm imagining an app would have a local store with an eventually-consistent view of the conversation. The app UI would only render the current state from that local store -- this lets you read when the device is offline, underground, etc.  So when the user opens a conversation, the app would issue a `Query` for the latest N messages in that conversation and begin streaming just that conversation's topic. Whenever the app received a streamed message it would optimistically add it to the local conversation state (so it shows up right away). Then the app just needs to eventually "sweep" with a `Query` which would pull that page of the conversation to make sure it picks up every message.

tl/dr: I think the app should never assume the stream is reliable. It's conceptually more like a lightweight push notification that the app can render immediately (optimistically) and then confirm later w/ a proper `Query` for that timespan of the conversation.

mkobetic:
@dmccartney, hard to disagree with anything you've said. However, I doubt any of the current users of the API are as thorough as you recommend and simply treat the all-stream as "sufficiently reliable". We've added the functionality based on popular demand, I'm just trying to address some of its limitations here.

mkobetic:
> @mkobetic Would have to confirm support in GRPC gateway and browser Fetch API

IIUC both of those should support HTTP2 transparently. I dug a bit around the LB situation and realized that given that we are using ALB (L7) the HTTP2 connection is not going to be a pass-through to the GW. So it's likely HTTP2 to LB and HTTP1 to GW atm. It might still be desirable to turn on HTTP2 for the second hop to gain some efficiencies, but it won't address the issue raised here. For that we'll really need to turn on sticky sessions https://github.com/xmtp-labs/infrastructure/pull/163

PR #31, opened 2022-12-14
Generate Kotlin Protobufs

nplasterer:
This adds Kotlin generated code for our protos.

The current protobufs for Kotlin require exporting to both --java_out and --kotlin_out. The Kotlin code calls into the Java code and creates separate Kotlin files with KT in the name. You can read more about this here: https://developers.google.com/protocol-buffers/docs/reference/kotlin-generated

We are also using [grpc-kotlin](https://github.com/grpc/grpc-kotlin/tree/master/compiler) and [grpc-java](https://github.com/grpc/grpc-java/blob/master/COMPILING.md).

We are currently releasing to github packages but in the future we should release to maven central as well.

I ran the publish gradle task and you can see the published proto here: https://github.com/xmtp/proto/packages/1754440


nplasterer:
The generated code for kotlin is rather messy. Not sure if we feel strongly about committing the generated files? Or if we should not commit them since the scripts will run the generate command anyways. I could go either way.

gradlew
```diff
@@ -0,0 +1,244 @@
+#!/bin/sh
```
neekolas, line 1: Does this file need to be in the root? Would be nice to have it live in the `dev/kotlin` folder just for the sake of uncluttering

nplasterer: Unfortunately yes. I tried moving it into the kotlin folder in these commits: [f16eced](https://github.com/xmtp/proto/pull/31/commits/f16eced15ef39a887c223b0190b28a3b42dddfdd) [9b8fbe2](https://github.com/xmtp/proto/pull/31/commits/9b8fbe2497aa8352d7b837b58aa8fc960870a903) but ran into errors with [gradle wrapper validation](https://github.com/gradle/wrapper-validation-action) which looks for the wrapper in the root.

neekolas:
I'm in favour of us `.gitignore`ing the generated code and leaving out of the repo. Otherwise this looks great (coming from a person who has an extremely fuzzy understanding of Gradle or Java)

gradlew
neekolas: OK. Not the end of the world

nplasterer: If we do move forward with a java package as well of the protos this will be reused between kotlin and java.

neekolas (approved):
I'm so glad someone other than me is banging their head against the wall trying to figure out how to make Java GRPC Codegen work.

Nice work @nplasterer!

mkobetic (approved):
I also know next to nothing about java or kotlin, but I don't see anything objectionable here. 👍

PR #32, opened 2022-12-16
Generate Java Protobufs

nplasterer:
This adds Java generated code for our protos. We don't plan on making a java sdk at this point in time but this can lay the ground work for if we plan to do so in the future. This was all kind of fresh since the kotlin protos required me to essentially make a java proto first.

You can read more about the java protobufs here: https://developers.google.com/protocol-buffers/docs/reference/java-generated

We are also using [grpc-java](https://github.com/grpc/grpc-java/blob/master/COMPILING.md).

We are currently releasing to github packages and I think this is fine since we don't plan to make a sdk but if we do make a java sdk someday we should probably move this to maven as well.

I honestly don't feel strongly about adding this or not to the proto repo. So if someone says it would be better to just close this I'd be okay with that.


nplasterer:
Since we don't plan on making a java sdk at this point in time I think I'm just going to close this out. This was all kind of fresh since the kotlin protos required me to essentially make a java proto first. But it's just going to add clutter to this repo if noone wants it currently. We can re-open this at a later date if java becomes a priority.

PR #33, opened 2022-12-19
Lock typescript type

nplasterer:
Fixed here: https://github.com/xmtp/proto/pull/30#discussion_r1033798195

Lets lock the typescript type so the tests don't fail


neekolas (approved):

mkobetic:
❤️ Thank you!

PR #34, opened 2023-01-04
feat: add a BatchQuery RPC method

michaelx11:
Github issue: https://github.com/xmtp-labs/hq/issues/899

This change introduces the concept of batch queries, which provide no-frills bundling of existing QueryRequests and QueryResponses.

Implementation details (parallelism, batch size limits) are omitted from the proto API and are left as implementation choices.


michaelx11:
Shoot, my proto version is slightly off which leads to a much larger diff set. I can lock and regen if that is preferable.

neekolas:
> Shoot, my proto version is slightly off which leads to a much larger diff set. I can lock and regen if that is preferable.

Yes please. I'm hoping to sneak in a little project to standardize our proto compiler into a Docker image some time in the next few months, but until then synchronizing versions is best.

neekolas (approved):

michaelx11:
@neekolas - I got the versions right but still a large diff. Some of those look expected, but do any of those diffs look out of place? The new line + // are a bit suspicious but not sure what causes those.

neekolas:
Odd that you are still getting diffs even with the versions matching, but they all seem related to an extra line of comments in the generated code, which doesn't matter.

:shipit:

michaelx11:
Ran e2e tests referencing `github.com/xmtp/proto/v3 v3.11.1-0.20230104230713-55e48609f0dd` and they passed, going to merge.

PR #35, opened 2023-01-06
Update Kotlin release versioning

nplasterer:
The previous version incrementing required two commits to update the version file. 
I believe this will get the version from the latest tag per these docs https://docs.github.com/en/actions/learn-github-actions/contexts.

### DIFF
--- a/.github/workflows/post_release.yml
+++ b/.github/workflows/post_release.yml
@@ -1,0 +1,25 @@
+name: Post Release
+on:
+  release:
+    types: [published]
+jobs:
+  kotlin:
+    name: Release (Kotlin)
+    runs-on: ubuntu-latest
+    steps:
+    - name: Checkout project sources
+      uses: actions/checkout@v3
+    - name: Setup Gradle
+      uses: gradle/gradle-build-action@v2
+    - name: Validate Gradle Wrapper
+      uses: gradle/wrapper-validation-action@v1
+    - name: Generate protobufs
+      run: dev/kotlin/generate
+    - name: Run build with Gradle Wrapper
+      run: ./gradlew build
+    - name: Gradle Publish
+      env:
+        RELEASE_VERSION: ${{ github.ref_name }}
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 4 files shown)


michaelx11 (approved):
`ref_name` should equal the tagged release right? LGTM

.github/workflows/post_release.yml
```diff
@@ -0,0 +1,25 @@
+name: Post Release
+on:
+  release:
+    types: [published]
```
neekolas, line 4: Nice

.github/workflows/release.yml
```diff
@@ -54,8 +54,7 @@ jobs:
       run: ./gradlew build
     - name: Gradle Publish
       env:
+        RELEASE_VERSION: ${{ github.ref_name }}
         GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         GITHUB_ACTOR: ${{ secrets.GITHUB_ACTOR }}
-      run: ./gradlew publish
-    - name: Increment version
-      run: ./gradlew incrementBuildVersion
+      run: ./gradlew publish
```
neekolas, line 60: Do we still need this if we are publishing in post-release?

neekolas (approved):

.github/workflows/release.yml
nplasterer: Yes because even though the publish has happened for the package we still need to publish the kotlin sdk which happens in this gradle task https://github.com/xmtp/proto/blob/main/kotlin/build.gradle#L64-L87 This is just calling this gradle task.

nplasterer:
@michaelx11 per the documentation
 
> For workflows triggered by pull_request, this is the pull request merge branch. For workflows triggered by release, this is the release tag created.

So I'm hopeful that this will be the tag name even though it is confusingly called a `ref_name`

.github/workflows/release.yml
nplasterer: O shoot now I see where your comment was I thought I had removed this entirely from the release.yml but yes you're right this entire kotlin block should be removed from here.

PR #36, opened 2023-01-18
[Kotlin] Fix post release GitHub action

nplasterer:
This removes the `post_release.yml` file that was added to trigger a workflow on release so that we could get the tag name for the kotlin package. It turns out per the github documentation that workflows cannot trigger other workflows when using the `GITHUB_TOKEN` this disallows unintentional recursive actions. See: https://docs.github.com/en/actions/using-workflows/triggering-a-workflow#triggering-a-workflow-from-a-workflow GitHub suggests using a Personal Access Token to get around this but I think that is less than ideal and a bit hacky.

There aren't a lot of great clean options then for us to get the tag for the release. I took some suggestions from this stackoverflow https://stackoverflow.com/questions/58177786/get-the-current-pushed-tag-in-github-actions. And decided to just get it from the git command. The only issue with this is that it will error if no version tag is present. But I think since this is triggered in the release flow the version tag should always be present?


.github/workflows/release.yml
```diff
@@ -38,3 +38,25 @@ jobs:
...
+    - name: Setup Gradle
+      uses: gradle/gradle-build-action@v2
+    - name: Validate Gradle Wrapper
+      uses: gradle/wrapper-validation-action@v1
+    - name: Generate protobufs
+      run: dev/kotlin/generate
+    - name: Run build with Gradle Wrapper
+      run: ./gradlew build
+    - name: Set Tag
+      run: echo "GIT_TAG=`echo $(git describe --tags --abbrev=0)`" >> $GITHUB_ENV
```
neekolas, line 56: Just to confirm. We want the `v` prefix (ie `v7.7.1`) for Gradle, right?

neekolas (approved):

PR #37, opened 2023-01-18
[Kotlin] Fix shallow repo fetch not getting tags

nplasterer:
3rd times a charm. When I ran the command locally it worked fine but in the github action it failed with 
```
Run TAG=`echo $(git describe --tags --abbrev=0)`
fatal: No names found, cannot describe anything.
```
From this [stackoverflow](https://stackoverflow.com/questions/4916492/git-describe-fails-with-fatal-no-names-found-cannot-describe-anything) I discovered that the github action was making a shallow clone of the repo that did not include the tag information. This should fix that. 🤞

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -44,7 +44,8 @@ jobs:
     steps:
     - name: Checkout project sources
       uses: actions/checkout@v3
-    - run: git fetch --prune --unshallow --tags
+      with:
+        fetch-depth: 0
     - name: Setup Gradle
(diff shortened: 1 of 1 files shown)


neekolas (approved):

nplasterer:
I feel like I'm shooting in the dark here a bit but from this updated document. It looks like this is the new way of getting all of the branches and tags https://github.com/actions/checkout 🤞 So changing to this since it's a bit cleaner.

PR #38, opened 2023-02-06
Keystore API protos

neekolas:
## Summary

After some discussion around https://github.com/xmtp/xmtp-js/pull/267 both in Github and [in Slack](https://xmtp-labs.slack.com/archives/C04HQ80PXMF/p1675294046814339), we have decided that it makes the most sense to wrap the Keystore API requests and responses in generated Protobuf types. It gives us greater optionality as we develop new Keystore implementations that require serialization (MM Snaps, possibly Rust) to be able to easily and consistently serialize the content as either JSON or a byte array.

It's worth noting that serialization is optional. For implementations like the InMemoryKeystore we would simply wrap requests/responses in the generated Protobuf types before calling the Keystore, but there is no need to actually serialize them.


proto/keystore_api/v1/keystore.proto
```diff
@@ -0,0 +1,106 @@
...
+    repeated Request requests = 1;
+}
+
+message CreateInviteRequest {
+    xmtp.message_contents.InvitationV1.Context context = 1;
+    xmtp.message_contents.SignedPublicKeyBundle recipient = 2;
+    uint64 createdAtNs = 3;
+}
+
+message CreateInviteResponse {
```
neekolas, line 89: I decided the right heuristic here is to have batch responses return errors in the response array and single-request endpoints simply throw.

package.json
```diff
@@ -62,6 +62,7 @@
   "dependencies": {
     "long": "^5.2.0",
     "protobufjs": "^7.0.0",
+    "rxjs": "^7.8.0",
```
neekolas, line 65: Seems this is required in the newer version of the JS proto compiler. I was getting errors building the TS types without

proto/keystore_api/v1/keystore.proto
```diff
@@ -0,0 +1,150 @@
...
+
+// Response to a CreateInviteRequest
+message CreateInviteResponse {
+    ConversationReference conversation = 1;
+    bytes payload = 2;
+}
+
+// Request to save a batch of invite messages to the Keystore
+message SaveInvitesRequest {
+    repeated xmtp.message_api.v1.Envelope envelopes = 1;
```
michaelx11, line 131: Echoing my outdated comment I made moments ago: https://github.com/xmtp/xmtp-js/pull/267/files#r1099023226

Can we exclude message_api protos from the Keystore API protos? It makes proto generation have more dependencies for Rust and doesn't add anything AFAICT. It'll look ugly to just have repeated byte arrays as this field, but I think it would work.

michaelx11 (approved):
Overall, awesome change

proto/keystore_api/v1/keystore.proto
neekolas: I'm alright with removing the dependency on `message_api`, since it's just used in one place. I wouldn't want to remove all the `message_contents` references though. We have a lot of those, and I think it's useful to be able to have consistent types across the SDK wherever possible. Re-implementing those feels like it would create a bunch of boilerplate, and new places for bugs to creep in.

neekolas: If our only concern is the GRPC dependency, happy to remove this and create a new type that just has the topic and the message.

michaelx11: No changes necessary then! If it's not an obvious move then I don't think it's worth cutting out message_api dependency

michaelx11: Oh wait sorry I misread your comment. Yeah we definitely don't need to replace `message_contents` references. Those feel natural and right to build upon.

jazzz (approved):

PR #39, opened 2023-02-17
feat: self-describing signature format

michaelx11:
**Goal**

Currently, our signature proto comprises serialized bytes and a recovery bit, but a couple critical pieces of context must be obtained out of band:
1) What is the message that was signed? Right now this is re-constructed via application code with out-of-band usage context e.g. this signature is for identity key association, so message = "XMTP : Create Identity..."
2) What is the encoding of the serialized signature bytes? What digest algorithm was used?

Our hope is to create a fully self-describing signature format that also includes explicit context to avoid cross-context signature re-use security issues. The concrete first objective is to support SIWE-like signatures for revocable send keys.

**Changes**

Add a 3rd signature type to the union: `Generic`

**Notes**

This initial PR is meant to power discussion. Let's iterate on everything here from naming to structure. Cheers!


proto/message_contents/signature.proto
```diff
@@ -22,8 +22,16 @@ message Signature {
         bytes bytes = 1; // compact representation [ R || S ], 64 bytes
         uint32 recovery = 2; // recovery bit
     }
+    // Self-describing signature format meant for maximizing interoperability
+    message Generic {
+        bytes signature_bytes = 1; // signature bytes
+        bytes message_bytes = 2;
+        string algorithm = 3; // signature algorithm
```
jazzz, line 29: Can you provide an example of values that would belong here?
Particularly interested in scope

proto/message_contents/signature.proto
```diff
@@ -22,8 +22,16 @@ message Signature {
...
+        string context = 4; // context the signature should be trusted for e.g. "sender_key_association"
```
jazzz, line 30: Interesting. What is the rationale behind tracking a context? Is there a particular threat you are mitigating?

proto/message_contents/signature.proto
```diff
@@ -22,8 +22,16 @@ message Signature {
         bytes bytes = 1; // compact representation [ R || S ], 64 bytes
         uint32 recovery = 2; // recovery bit
     }
+    // Self-describing signature format meant for maximizing interoperability
+    message Generic {
+        bytes signature_bytes = 1; // signature bytes
+        bytes message_bytes = 2;
```
neekolas, line 28: In what cases would the `message_bytes` be full content and in what cases would it be a digest?

For V2 messages, we use currently sign the digest of `concat(headerBytes, messageContents)`. If the expectation was to use the raw inputs instead of the digest, we would effectively be doubling the size of our messages.

proto/message_contents/signature.proto
neekolas, line 30: It feels like `context` should be an enum here, with a clearly defined and documented set of contexts.

proto/message_contents/signature.proto
neekolas, line 29: This also feels like it should be an enum. We can always add values as we add more signature algorithms.

proto/message_contents/signature.proto
jazzz: Great point. 

Seems like Association Signature, and DigestSignature should be handled differently. From the lens of 'self describing' a signature to prove integrity needs the hashing algorithm and approach (eg. EIP 191)

Where Signatures used to prove user approval need extra context such as the text that was submitted. 

Toy Example:
```
AssociationProof :{
    Message,
    Signature: {
          type: ECDSA_keccak256_eip191,
          bytes:  ...
    }
}
```

proto/message_contents/signature.proto
michaelx11: agreed!

proto/message_contents/signature.proto
michaelx11: If we carry the signed message payload without context, it'll be easy to accidentally do something along the lines of:
```
is_ok = public_key.verify(proto.message_bytes, proto.signature_bytes)
```
in our code unless we're very clear about what context the signature should be used for. This happens implicitly now because we need to regenerate the message using out-of-band context during signature verification.

proto/message_contents/signature.proto
michaelx11: If we carry the signed message payload without context, it'll be easy to accidentally do something along the lines of:
```
is_ok = public_key.verify(proto.message_bytes, proto.signature_bytes)
```
in our code unless we're very clear about what context the signature should be used for. This happens implicitly now because we need to regenerate the message using out-of-band context during signature verification.

michaelx11: The alternative is we define one union member per context, so we would add multiple signatures in the future:
```
    oneof union {
        ECDSACompact ecdsa_compact = 1;
        WalletECDSACompact wallet_ecdsa_compact = 2;
        SIWESendKeyProof siwe_send_key_proof = 3;
        RevokeSendKeyProof revoke_send_key_proof = 4;
    }
```

I'm not opposed to doing this actually, it's probably the most explicit and clear way to structure this @jazzz @neekolas

michaelx11: In that sense we already have the framework for a self-describing signature proto, the `oneof union` might be all the abstraction we need.

michaelx11:
From our discussion I see a few things that need to be captured/enumerated in our proto, totally open on how to design it:

1) [SCHEME] Information required to successfully verify the signature
  a) Original message being signed
  b) Digital signature algorithm
  c) Any processing and digest scheme used to reduce the original message into an input into the digital signature algorithm
2) [PURPOSE] Purpose of the signature (almost like a [domain separator](https://ethereum.stackexchange.com/a/125129)) which ensures a signature can only be used for its prescribed purpose
  a) Note: currently this is not encoded in the proto in any way, instead it is implicitly derived from the parent protos containing`Signature` protos and the exact use case within XMTP application logic
  b) Examples: `wallet_to_identity_key_association`, `wallet_to_send_key_association`

Two possible schemes:
1) As above, overload the `oneof union ...` with a new `Signature` message type for every (scheme, purpose) pair. Results in a cross-product and lack of re-use of similar signature formats
2) Define a new message type for `oneof union ...` for every scheme. Have a separate enum for every `purpose` and include it in `Signature` message types as a `context` field.

Thoughts? @jazzz @richardhuaaa @neekolas

proto/message_contents/signature.proto
michaelx11: Yup, that's the current proposal! The idea is to make sure we have enough info in the proto to verify the message e.g. is this SIWE for my address? Is it for XMTP and not some other use case?

proto/message_contents/signature.proto
michaelx11: Is the proposal to create a new proto at a higher level for "proof"? That could bundle the context and the signature, separating signature math specifics from the use case and message verification.

jazzz:
I think this change might need to be more narrowly focused. I've learned from this discussion that the term Signature is an overloaded term in XMTP and used in many conflicting ways. A single solution may not be the most appropriate at this time. 

Rather than focusing on All signatures used across the XMTP protocol suite, I think it might help to focus on one of these applications, the Association Signature. (Since signature is overloaded I'm going to propose the term "Proof" is used, aka the AssociationProof.)

I think the most immediate work needed is to relax the definition of the AssociationProof. Where Signatures used for message contents could remain unchanged for now.

michaelx11:
Got it, so immediate work is to make proto and code changes to allow an `AssociationProof` to stand in for what is currently accomplished via `Signature.WalletECDSACompact` and `xmtp-js` application logic?

jazzz:
> Note: the context (aka domain aka purpose) would need to be part of the signed message.... before actually passing it into the signature algorithm.

I like the intention however I see some issues with the current formulation. If the intention is to allow developers more flexibility in which signatures are used in adding this requirement would have the opposite effect. e.g. Using SIWE with a ContextPrefix would make the the Signature non compliant with https://eips.ethereum.org/EIPS/eip-4361. Furthermore achieve this this a EIP191 compliant wallet would mean that the signature text would need to be injected into the text (which would force app developers to change their signature texts in a destructive way.

I think there is a way to achieve this though:
In the case of the existing AssociationProof, The textPrefix "Create Identity" embeds the context in the signature.
In the case of SIWE this is achievable via controlling the resource line, and embedding the context there.

I think its fair to say that all supported AssociationProof types MUST have an embedded context to guard against signatures being used out of context.

(Sidenote: The 'Context' approach I think would be highly applicable when reformatting digestSignatures, which are signed by XMTP keypairs and are not externally/user facing)

jazzz:
> Got it, so immediate work is to make proto and code changes to allow an `AssociationProof` to stand in for what is currently accomplished via `Signature.WalletECDSACompact` and `xmtp-js` application logic?

Yes. and the MessageContentProof could be address separately using an approach unique to its specific needs

michaelx11:
Closing in favor of: https://github.com/xmtp/proto/pull/43

PR #40, opened 2023-02-21
OSSRH Kotlin Package Release

nplasterer:
Following the [Open Source Repository Hosting documention](https://central.sonatype.org/publish/requirements) to add additional information to the package release. 
Kotlin will now be released as a GitHub Package and a Maven Central Repository Package to allow easier access from the consuming integrations. In the future we may be able to delete the releasing to GitHub Packages.


neekolas (approved):
I have no idea how Gradle works, but approving to unblock

PR #42, opened 2023-02-22
fix: kotlin OSSRH needs gpg signing

nplasterer:
Trying to test this locally before releasing again.
https://central.sonatype.org/publish/publish-gradle

https://github.com/xmtp/proto/actions/runs/4246666696/jobs/7383760942


jazzz (approved):
It's all ~greek~ gradle to me. :shipit:

PR #45, opened 2023-02-23
fix: Update group id match OSSHR

nplasterer:
Per this conversation https://issues.sonatype.org/browse/OSSRH-89132?page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel&focusedId=1244733#comment-1244733

> Profile target mismatch 

Means that the groupID was not lining up correctly this aims to correct that.


elisealix22 (approved):
Wow, great find!

PR #47, opened 2023-02-27
Update topic map

neekolas:
## Summary

Oof. I realized that the `InvitationV1` doesn't include a `created` timestamp, which I need to sort the data


neekolas:
Thanks. 🤦

PR #48, opened 2023-02-28
Association proof + send key protos

richardhuaaa:
Outcome of previous discussion from what I understand:
1. We are sticking to an association proof type rather than folding it into the layer above
2. Wait for the send key changes before landing the association proof changes - @michaelx11 asked me to go ahead and put up a PR that includes both

Tbh I don't think the send key changes are too disruptive! This is basically the same as the previous PR (#43), but the changes to public_key.proto are different. I'll leave some explanatory comments below.

### DIFF
--- a/go/message_contents/public_key.pb.go
+++ b/go/message_contents/public_key.pb.go
@@ -213,8 +213,8 @@ func (x *SignedPublicKeyRevocationV1) GetProof() *RevocationProof {
 	return nil
 }
 
-// Wrapper for send keys. There will likely be additional metadata in the future,
-// for example app/device label
+// Wrapper for send keys. There will likely be additional metadata in the
+// future, for example app/device label
 type SendKey struct {
 	state         protoimpl.MessageState
 	sizeCache     protoimpl.SizeCache


proto/message_contents/proof.proto
```diff
@@ -0,0 +1,32 @@
...
+import "message_contents/signature.proto";
+
+option go_package = "github.com/xmtp/proto/v3/go/message_contents";
+option java_package = "org.xmtp.proto.message.contents";
+
+// Association Proof proves that identity key material is associated with
+// a specific individually.
+message AssociationProof {
+    // Legacy: "XMTP : Create Identity ..." Signature Payload
+    message LegacyCreateIdentityProof {
```
jazzz, line 15: I was originally hesitant about making these Messages over enums, however in the future messages will make it easier to support 1271

proto/message_contents/public_key.proto
```diff
@@ -26,6 +27,25 @@ message UnsignedPublicKey {
...
 
+message SignedPublicKeyV2 {
+    bytes key_bytes = 1;  // embeds an UnsignedPublicKey
+    AssociationProof proof = 2; // signs key_bytes
+}
+
+message SignedPublicKeyRevocation {
+    bytes key_bytes = 1;  // embeds an UnsignedPublicKey
+    AssociationProof proof = 2; // signs key_bytes
+}
```
jazzz, line 38: I'm uncertain that Revocation messages ought to support dynamic signatures.

Allowing for a association proof could provide an DOS attack surface, as a malicious app could attempt to hide a READKey revocation message in a SIWE message for example. Is this something that should be supported?

Requiring a explicit message "REVOKE XMTP KEY 0x11111...." would provide a clear visual cue about what is occurring. This would however make it difficult to revoke multiple keys at once. Thoughts?

proto/message_contents/proof.proto
```diff
@@ -0,0 +1,32 @@
...
+        // Text is the identitySignaturePayload e.g. Create Identity ...
+        bytes text = 1;
+        // Signature is the same as SignedPublicKey.Signature historically
+        Signature signature = 2;
```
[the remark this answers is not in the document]
jazzz: > are the same thing right (effectively a TextSchema)? 

No. in my mind they are different -- however the names could be be improved.

SignatureValidation -> TextSchema
SignatureGeneration -> what spec was used to generate the signature. e.g EIP191, CAIP-X

One addresses which text was passed to the signer. the other addresses which specification the signer uses to create signatures.

proto/message_contents/proof.proto
```diff
@@ -0,0 +1,32 @@
...
+package xmtp.message_contents;
+
+import "message_contents/signature.proto";
+
+option go_package = "github.com/xmtp/proto/v3/go/message_contents";
+option java_package = "org.xmtp.proto.message.contents";
+
+// Association Proof proves that identity key material is associated with
+// a specific individually.
+message AssociationProof {
```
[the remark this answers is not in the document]
jazzz: I don't quite follow the goal.

If there are multiple resources being used in the SIWE, (e.g. XMTP-CREATE-IDENTITY, XMTP-PROVISION-SENDKEY)

Where would the signature and text be stored if not attached to the PublicKey?

proto/message_contents/public_key.proto
```diff
@@ -26,6 +27,25 @@ message UnsignedPublicKey {
     }
 }
 
+message SignedPublicKeyV2 {
+    bytes key_bytes = 1;  // embeds an UnsignedPublicKey
+    AssociationProof proof = 2; // signs key_bytes
+}
```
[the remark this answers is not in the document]
jazzz: > one for new contact bundles (with identity key and pre-key)

Is there a direct need to generate a new ReadKey? or could the existing IdentityKey be used for the time being? -- That is clients search for readKeys in the ContactTopic, and use the existing identityKey. 

It would be great to keep IdentityKey rotation (which is a parallel work stream)  separate if possible.

proto/message_contents/proof.proto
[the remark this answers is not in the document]
jazzz: I don't immediately see how I would get the shared proof as a recipient in order to validate a key.

The message sent contains a SignedPublicKey, how would I verify the key was valid?

proto/message_contents/public_key.proto
jazzz: 1) Heres an example of the SIWE which uses resources https://docs.login.xyz/general-information/siwe-overview/eip-4361#example-message-to-be-signed

2)  `isn't this symmetric between provisioning a new key, and revoking an old key?` It could be made to be symmetric.

My best argument for asymmetric process would be; If a malicious developer phishes a key by generating an 'Underhanded' SIWE, a user can revoke that key and all messages would be flagged as potentially spam.

If a malicious developer phishes a revocation, the user has no recourse. Those messages will be flagged in all other recipients. Since revocation is permanent, having an explicit revocation text (rather than a SIWE) would make it clearer to users the impact of this action.

examples:

```
AcmeApp.org wants you to sign in with your Ethereum account:
0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2

URI: https://AcmeApp.org/login
Version: 1
Chain ID: 1
Nonce: 32891756
Issued At: 2023-02-28T01:25:24Z
Resources:
- ipfs://bafybeiemxf5abjwjbikoz4mc3a3dla6ual3jsgpdr4cjr3oz3evfyavhwq/
- xmtp://revoke-key/b8bab08eb3e9301a430a4104a39004b1a57e6f6fd0f6f460f7fc805415fa31e1c9515940612486dc45ce06148ce5c15e14f3cf83a5f8114ada66a641ca30ba510168a58a2cdbe7289c584f42
``` 

```
==========================================
XMTP REVOCATION REQUEST
==========================================
Signing this request will invalidate the associated key, and mark all messages as invalid. This cannot be undone. For more information visit: 'https://xmtp.org/KeyRevocation' for more information 

KeyDetails:

createdAt: 2023-02-21
key:b8bab08eb3e9301a430a4104a39004b1a57e6f6fd0f6f460f7fc805415fa31e1c9515940612486dc45ce06148ce5c15e14f3cf83a5f8114ada66a641ca30ba510168a58a2cdbe7289c584f42
```

michaelx11 (commented):
Looks good overall, we should use a different Proof type for each use case i.e. AssociationProof = these bytes <> this identity, and RevocationProof = these bytes </> this identity.

proto/message_contents/proof.proto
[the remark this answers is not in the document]
michaelx11: I see, so one way of kind of formulating the proposal above is to have:

1) Different use case proofs (Association, Revocation)
2) Different types of proofs for each use case (AssociationLegacy, AssociationSIWE)
3) Everything below this level is either a Signature or a protobuf primitive (AssociationLegacy.signature, AssociationLegacy.bytes, ...)

I generally agree with that, but with the caveat that I think having `SignatureSpec` and `SignatureFormat` allows us to reuse code more within layer (2)), but I don't have too strong feelings. As long as we don't have the one layer approach of:

```
AssociationProof {
...
every
field
ever needed by any variation
...
}
```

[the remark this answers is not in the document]
jazzz: I'm not too picky here and don't want to block development. -- More important that we have a robust versioning system which can incorporate new types in the future.

The core reason why I would push for "SignatureGeneration" method be captured somewhere -- perhaps the signature is the correct place. Is to capture different levels of responsibility.

SignatureFormat -> Determined by blockchain
SignatureGeneration -> Determined by wallet developers
SignatureValidation -> Determined by app developer

By coupling two of these together future development in complicated -- Particularly as XMTP tries to become/remain chain agnostic an implicit assumption that all signatures are EIP191 will mean developers are excluded unnecessarily.

proto/message_contents/public_key.proto
```diff
@@ -26,6 +27,25 @@ message UnsignedPublicKey {
     }
 }
 
+message SignedPublicKeyV2 {
+    bytes key_bytes = 1;  // embeds an UnsignedPublicKey
```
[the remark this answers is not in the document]
michaelx11: Oh yup, you're right. I forgot this is an embed of the UnsignedPublicKey

proto/message_contents/public_key.proto
```diff
@@ -26,6 +27,25 @@ message UnsignedPublicKey {
...
+    AssociationProof proof = 2; // signs key_bytes
+}
+
+message SignedPublicKeyRevocation {
+    bytes key_bytes = 1;  // embeds an UnsignedPublicKey
+    AssociationProof proof = 2; // signs key_bytes
```
michaelx11: Given our talk I thought this would be a RevocationProof message? The idea being the proof's type makes it super clear what the purpose of the proof is. Otherwise, is there a distinguishing factor between an AssociationProof used to associate and one used to revoke?

proto/message_contents/proof.proto
```diff
@@ -0,0 +1,46 @@
...
+}
+
+// A revocation proof proves that a child key (and/or other data) has been
+// revoked by a parent key
+message RevocationProof {
+    // A custom payload format defined in a future RFC
+    message SignedRevocationRequest {
+        bytes text = 1;
+        Signature signature = 2;
+    }
```
jazzz, line 41: how will a developer know how this text ought to be validated? 
Is the intention that it is static for this message type? or is dynamic?

proto/message_contents/proof.proto
```diff
@@ -0,0 +1,46 @@
...
+
+    oneof union {
+        SignedRevocationRequest signed_revocation_request = 1;
```
jazzz, line 44: if this union is to support versions can the SignedRevocationRequest be primitively gvie a version number so its consistent in the future?

proto/message_contents/proof.proto
```diff
@@ -0,0 +1,46 @@
...
+import "message_contents/signature.proto";
+
+option go_package = "github.com/xmtp/proto/v3/go/message_contents";
+option java_package = "org.xmtp.proto.message.contents";
+
+// An association proof proves a child key (and/or other data) has been
+// associated with a parent key.
+message AssociationProof {
+    // Legacy: "XMTP : Create Identity ..." format
+    message LegacyCreateIdentityProof {
```
jazzz, line 15: DUST: Legacy will have less and less meaning and time moves on. 

Rename to 'CreateIdentityProof' or 'StaticCreateIdentityProof' ?

proto/message_contents/public_key.proto
michaelx11: oh I must have been on a super old version of this file, disregard

proto/message_contents/public_key.proto
```diff
@@ -26,6 +27,30 @@ message UnsignedPublicKey {
     }
 }
 
+// Signed public key that uses self-describing signatures (aka
+// 'AssociationProof')
+message SignedPublicKeyV2 {
+    bytes key_bytes = 1;  // embeds an UnsignedPublicKey
```
jazzz, line 33: Clarifying: Its an explicit choice to have the bytes being associated with a wallet be external to the associationProof?

jazzz (commented):
I've added a some last comments (mostly DUST). I'm happy to move out of the way to keep this moving. 

Summary: 
- Adding verbosity to names would help, new developers understand what it going on.
- Added explicit V1's to names so its consistent and clear how future versions should be added

michaelx11 (commented):
Awesome work, looks super close. We can refine the exact fields and stuff later but definitely have a good structure down

proto/message_contents/proof.proto
michaelx11: This is a great point

jazzz (approved):
giving this a 👍 as I don't have any hard blockers. I think there are some fixups but @richardhuaaa can decide what to keep and what to leave behind.

proto/message_contents/public_key.proto
jazzz: for simplicity I've tried to break down the usages of signatures into different concepts which allows them to be specialized. AssociationProof only addresses a wallet looking to associate a key with its account.

I Imagine signing a message for integrity would be captured in an IntegrityProof which would have different requirements -- Favoring a digest over the input text for example

michaelx11 (approved):
ship it, the overall structure seems good and we can add more fields to specific Proof submessages as we need them

PR #49, opened 2023-03-02
Send key proto tweaks

richardhuaaa:
The generated typescript definitions for these protos don't use the name of the `oneof` field, and directly nest the children as properties. Using more descriptive names for the children in light of this.

Also, Send Keys can use proofs from the outset, so not listing SignedPublicKey v1 as an option.


jazzz (approved):
🚀

PR #50, opened 2023-03-02
Update signed private key to support proofs

richardhuaaa:
I would prefer to convert this into a `oneof`, but I can't confirm with certainty that this will work across all languages. I need to use the [`generate` method](https://github.com/xmtp/xmtp-js/blob/8cae820b247a2bf7663c277c2d3a2e03f02493b4/src/crypto/PrivateKey.ts#L48) of SignedPrivateKey to create send keys, for now I'll modify that method to serialize in both formats.


jazzz:
Can you explain the necessity of this change in more detail?

- > but I can't confirm with certainty that this will work across all languages.

   - Which languages are there concerns around? What is the source of the potential incompatibility? 

- SignedPrivateKeys are not in in the SDK's at this time. Does this change how the problem should be approached at all?

michaelx11:
Do we need to change `SignedPrivateKey` at all or can we create a new `SendKey` type? For backwards compatibility (V3 -> V2), we can just take the SendKey's `SignedPublicKeyV2` fields and massage them into a `SignedPublicKey`. The constructed `SignedPublicKey` can be used in message headers, invites, etc

michaelx11:
There's already a good amount of duplication between `SignedPrivateKey` and `PrivateKey`.

There is currently this nice mapping between protos and typescript classes right now, one Typescript class to implement one proto message. A change like this kind of commits us to shoving all SendKey private key logic into `SignedPrivateKey` going forward. It'll live side by side with all the old code. Not necessarily a bad approach but things could get messy versus a completely separate class.

From what I'm reading we need `generate` to create new SendKey private material. The problem is that this returns a `SignedPrivateKey (old)` type with an old SignedPublicKey? One approach is to have a SendKeyPrivate class that extends `SignedPrivateKey` and just wraps `generate` and does some shuffling to create a SignedPublicKeyV2 from the SignedPublicKey.

jazzz:
> I can see them here: https://github.com/xmtp/xmtp-js/blob/8cae820b247a2bf7663c277c2d3a2e03f02493b4/src/crypto/PrivateKey.ts#L28. 

I should have been more precise. No clients are generating SignedPrivateKeyBundles, and to my knowledge they are not persisted on the network at all. This is the last step of the SignedKey migration path is to enable users to generate these keyTypes. From a lens of backwards compatibility, I'm not sure the existing keyType needs to be supported.

> I don't think there is anything different about a send key private key compared to any other private key. 

Once full deployed SendKeyBundles cannot be used where SignedPrivateKeyBundles can. So having some ability for clients to discern between them will be needed -- thought its can occur during key discovery via topics

michaelx11:
Dumb question: why can't SendKey be serialized with an `AssociationProof` if there isn't a `SignedPublicKeyV2` in the `SignedPrivateKey`? I know you have a link I just can't find it off the top of my head.

I think this confusion is why I'm trying to push for having new types for anything related to the SendKey effort as much as possible. If we need to worry about legacy, we can just stick those bytes from the new protos into the old format

jazzz:
Where in the flow are sendKey-PrivateKeys being serialized? My understanding was that SendKeys were largely ephemeral.
If they are not materialized and the intention is to convert the Public and private components into the legacy types if needed. I'm confused as to why there is a need for changes to the stored PrivateKey.

michaelx11:
I don't think I fully understand the coupling. Could you give the me the TL;DR again? Basically, gotta disprove my current thinking which is:

```
[v2 clients] (could be locked today) <--- conversion code ---> [v3 clients, with all new proto types for signedpublic and signedprivate keys]
```

PR #51, opened 2023-03-03
Add peer_address

neekolas:
## Summary

Adds a `peer_address` field to simplify management of conversations. This allows the `ConversationReference` to be free-standing, and all the SDK needs to work with a conversation.

## Notes
Because these protos are not currently persisted anywhere, I can add new fields out of order without worrying about breaking anything.


michaelx11 (approved):
It does hurt my soul a bit, but like you said no practical implications for reordering fields. Ship it!

neekolas:
I know. It's everything I've been taught not to do with Protobuf. 

But I'd rather have them in a pretty order forever.

PR #52, opened 2023-03-09
New signature and send key protos

richardhuaaa:
Another iteration of protos for generic proof types and send keys incorporating lessons from a prototype I have built and prior discussion. Hopefully this is much cleaner, look forward to the discussion!

Previous discussion: https://github.com/xmtp/proto/pull/48


proto/message_contents/signature.proto
```diff
@@ -27,3 +30,59 @@ message Signature {
...
+        Signature signature = 2;
+    }
+    oneof version {
+        V1 v1 = 1;
+    }
+}
+
+// Dynamic signature using sign-in with Ethereum format (EIP-4361)
+// The use-case is inferred from the signature text
+// Issue: use-case might be ambiguous
```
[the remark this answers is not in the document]
michaelx11: I'm down for a ClaimType field!

proto/message_contents/public_key.proto
```diff
@@ -27,50 +26,58 @@ message UnsignedPublicKey {
     }
 }
 
-// Signed public key that uses self-describing signatures (aka
```
mkobetic, line 30: I assume these were all added in the previous prototype and so we're not worried about backward compatibility with these?

mkobetic:
I'm not quite up to speed on the send key / account linked key ideas so can't really comment on that, but I don't see a structural issue here.

One more general remark, the ambition was that between PROTOCOL.md and the proto files we should have a fairly exhaustive definition of the protocol, ideally sufficient for someone to implement it themselves. That would probably help someone like me to ponder the semantics as well.

Also at some point we should start getting into the habit of writing XIPs for these changes.

proto/message_contents/private_key.proto
```diff
@@ -12,7 +12,34 @@ import "message_contents/public_key.proto";
 option go_package = "github.com/xmtp/proto/v3/go/message_contents";
 option java_package = "org.xmtp.proto.message.contents";
 
-// PrivateKey generalized to support different key types
+// A private key with a corresponding public key that is provably linked
+// to a user account
```
mkobetic, line 16: What makes the key account linked?

jazzz: The "Create Identity Signature" is used to Link a blockchain account to an XMTP Account/Identity Key. This is the definition of account linking. The concept behind an accountLinkedKey is to split up how keys are used in the Protocol by intent rather than cryptographic construct. 

E.g The sendKey work will enable a wider range of signatures to be used to link accounts such as SIWE

mkobetic: Ah right, I confused myself. SignedPrivateKey isn't actually signed itself, it's the corresponding PublicKey that is signed. It's been a while 😓

proto/message_contents/invitation.proto
```diff
@@ -11,6 +11,73 @@ import "message_contents/public_key.proto";
 option go_package = "github.com/xmtp/proto/v3/go/message_contents";
 option java_package = "org.xmtp.proto.message.contents";
 
+// Unsealed invitation V2
+message InvitationV2 {
+    // Supported encryption schemes
+    // AES256-GCM-HKDF-SHA256
```
jazzz, line 17: Is the intention to use this Invitation for Invites sent to participants? Or only the invites sent to yourself?

proto/message_contents/invitation.proto
```diff
@@ -11,6 +11,73 @@ import "message_contents/public_key.proto";
...
+    // message encryption scheme and keys for this conversation.
+    oneof encryption {
+        // Specify the encryption method to process the key material properly.
+        Aes256gcmHkdfsha256 aes256_gcm_hkdf_sha256 = 3;
+    }
+
+    // The address of the person being invited to the conversation. This
+    // field must only be used if the send_key_bundle attached
+    // to the invitation header originates from your own account.
+    string recipient_address_untrusted = 4;
```
jazzz, line 47: As I read this, this field is encrypted on the network so it is private, and unsigned so it is deniable. Is this true?

proto/message_contents/private_key.proto
```diff
@@ -65,13 +84,29 @@ message PrivateKeyBundleV1 {
...
+    repeated SignedPrivateKey pre_keys = 2;
+}
 
 
 // Versioned PrivateKeyBundle
 message PrivateKeyBundle {
     oneof version {
         PrivateKeyBundleV1 v1 = 1;
         PrivateKeyBundleV2 v2 = 2;
+        PrivateKeyBundleV3 v3 = 3;
```
jazzz, line 109: 🔥

proto/message_contents/signature.proto
```diff
@@ -27,3 +30,59 @@ message Signature {
...
+}
+
+// A signature used to link a key to an account, with
+// sign-in with Ethereum format (EIP-4361).
+// The key_bytes and role must be validated
+// against the signature text
+message AccountLinkedSIWESignature {
+    // V1
+    message V1 {
+        AccountLinkedRole role = 1;
```
jazzz, line 81: Can you explain how this mechanism works?

If the intention is to use SIW* resource lines to define keys. Isn't the relationship between Signature and Role 1:M?

Is there a need that this is captured when persisted rather than determined during validation? i.e. a Signature is validated for a specific role at runtime? e.g.`signature.validate(inputBytes, ROLE_SEND) -> false`

michaelx11 (approved):
Looks good!

jazzz (approved):
I like where this ended up. Looks good for Beta

PR #55, opened 2023-03-11
Add CODEOWNERS and triage action

nplasterer:
This adds a codeowner file and a triage action for new issues reported.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,0 +1,3 @@
+# Global rule:
+*     @xmtp/engineering
+*.md  @jhaaaa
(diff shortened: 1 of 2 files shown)


.github/CODEOWNERS
```diff
@@ -0,0 +1,3 @@
+# Global rule:
+*     @xmtp/engineering
```
nplasterer, line 2: I figure since this spans a lot of engineers it might be okay to use engineering here. But also could be convinced to use a smaller subset of proto pod people if we want to create a team?

jazzz (approved):

.github/CODEOWNERS
jazzz: Seems good for now, given its unclear exactly who should be on that list. Once it becomes clear the list can be tightened down

PR #57, opened 2023-03-14
Add SignRequest

neekolas:
## Summary

- Adds new Keystore request type `SignRequest` used to sign an arbitrary digest with either the `IdentityKey` or a selected `preKey`. Required for V2 message signing.
- Adds support for serializing `DecodedMessage` instances


proto/keystore_api/v1/keystore.proto
```diff
@@ -161,6 +164,16 @@ message CreateAuthTokenRequest {
     optional uint64 timestamp_ns = 1;
 }
 
+
+// SignRequest is used to sign a digest with either the identity key or a prekey
+message SignRequest {
+    bytes digest = 1;
```
michaelx11, line 170: I don't love how generic the SignRequest is. My (currently unfounded) fear being that identity_key signatures and even pre_key signatures can potentially have multiple applications in the future. I'm paranoid about potential cross-usage confusion if all we supply is a digest.

I wouldn't mind a separate SignRequest submessage for each intended application

neekolas: So, I can kinda get my head around breaking up the "sign with identity key" and "sign with preKey" usages into separate APIs.

Are you imagining anything more granular than that? They would both still take digests as an argument and return `xmtp.message_contents.signature.Signature` as a response.

michaelx11: Nope, just use-case level granularity is perfect IMO

neekolas: I hear you on the explicitness of that approach. It is a lot more data going over the API boundary, though. From a 32 byte digest to a potentially 1mb message. Are we sure it's worth the performance cost?

michaelx11: That is definitely a legit concern. I guess we could separate into "messageContentSignRequest" and "prekeySignRequest" if we're feeling spicy. I'm also okay with doing this simple generic approach for now and refining it later! It's just hard to get the cat back in the bag once it's out i.e. we'll need to support generic signatures for a while until we can fully deprecate.

I also want to get a second opinion to make sure I'm not overcomplicating: @richardhuaaa @jazzz

mkobetic: Why not just have generic digest, a purpose code and sign over both?

neekolas: @mkobetic this is moving some existing signing functionality to use the keystore (V2 message signing specifically). Changing the data that gets signed would require a complicated migration, so I think we should rule that out as an option for now.

mkobetic: OK, I'm probably just not understanding the discussion here. My concern was proliferation of protocol messages with identical structure.

jazzz: @michaelx11 if a signature specification read `all signatures used in xmtp MUST be locked to a specific context` would that address the majority of the concern? The idea being that in the future all signatures must be explicit about where they can be used

neekolas: That makes sense to me. I'll rename this proto and the endpoint to make space for a contextual signature endpoint that will come later. 

`signDigest` as the endpoint name?

michaelx11: Sounds good to me!

neekolas: I don't see this changing for V2 message payload signatures. If we want to introduce new signature types, we would need new message types since the change would not be backwards compatible.

michaelx11 (approved):
LGTM with the generic SignDigestRequest

PR #58, opened 2023-03-17
SignedEciesCiphertext

neekolas:
## Summary

We would like to add a signature to the ECIES payloads used for local conversation caching in the Keystore before release. The concern is that an attacker would be able to create an ephemeral private key and write valid entries into the user's local conversation cache. The SDK would then trust these entries, as the decryption would succeed so long as the user's public key was the recipient in the attacker's DH.

An exploit would require the attacker to already have access to the application context, but given that LocalStorage is not protected this seems possible (for example, a script injected into the page could hijack LocalStorage and create maliciously crafted conversations).

With the addition of the signature, I am also defining a protobuf type for the ECIES ciphertext, which was previously serialized manually.

## Issue
https://github.com/xmtp/xmtp-js/issues/295


proto/message_contents/ciphertext.proto
```diff
@@ -20,8 +22,24 @@ message Ciphertext {
...
+        bytes ephemeral_public_key = 1;  // 65 bytes
+        bytes iv = 2;                    // 16 bytes
+        bytes mac = 3;                   // 32 bytes
+        bytes ciphertext = 4;  // encrypted payload with block size of 16
+    }
+
+    // serialized Ecies message
+    bytes ecies_bytes = 1;
+    // signature of the ecies_bytes
+    xmtp.message_contents.Signature signature = 2;
```
michaelx11, line 44: Looks good, we should document the digest algorithm used for this (sha256, sha3, etc). I've been marching around with my "no signatures without context" banner for a while, but I think those changes can be bundled in later. We're exploring Proofs at the moment with the SendKey work.

michaelx11 (approved):

proto/message_contents/ciphertext.proto
jazzz: how exactly will this be signed? Wallet Signature?

neekolas: Good call out. I'll add some clearer docs. 

This is an IdentityKey signature, since it needs to happen in the background (new signature happens every time a conversation is added to the store). 

I thought about using preKeys, but then we would need to include metadata to tell you which key to verify with.

neekolas: `preKey` is an option. We'd just have to reverse out the public key from the signature and find the right one

PR #61, opened 2023-03-27
fix: dynamically update the pubspec version to match the release

elisealix22:
The last dart proto publish failed because the pubspec file couldn't be found. It also wasn't using the correct version. This PR attempts to update the version to the latest tag during the GitHub `release` Action to fix the publish.

Ref: https://dart.dev/tools/pub/automated-publishing#triggering-automated-publishing-from-github-actions
Failed Action: https://github.com/xmtp/proto/actions/runs/4535405973/jobs/7990814265#step:8:9

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -39,8 +39,16 @@ jobs:
       - run: dev/dart/generate
       - run: dart test
         working-directory: ./dart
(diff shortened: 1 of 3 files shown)


nplasterer (approved):
Worth a try!

PR #66, opened 2023-03-29
fix: listen for tags to publish dart

elisealix22:
Unfortunately, dart requires a tag to kick off the GitHub Action in order to publish the release. The last build failed because it was triggered by a push to `main` even though it had the correct tag and version in the `pubspec.yaml`: https://github.com/xmtp/proto/actions/runs/4544664990/jobs/8010997528#step:10:81

This is an attempt to listen for tags created by the JS release process. If the creation of the tag as a part of the JS release flow does not trigger this action, then I believe we will have to manually push a specific tag like `vX.Y.Z-dart` to trigger a publish.

### DIFF
--- a/.github/workflows/publish-dart.yml
+++ b/.github/workflows/publish-dart.yml
@@ -1,0 +1,39 @@
+# .github/workflows/publish.yml
+name: Publish to pub.dev
+on:
+  push:
+    tags:
+    # must align with the tag-pattern configured on pub.dev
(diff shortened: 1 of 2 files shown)


nplasterer (approved):
If this doesn't work is there anyway we can manually publish dart instead? Or does it have to be triggered from a tag. If we have to do it manually I think it would be cleaner if it was just a manual script or something. But don't feel strongly.

PR #71, opened 2023-03-30
Fix Go Generation

neekolas:
## Summary

@nakajima realized that while we were generating Go code with the latest round of Keystore changes, it wasn't actually valid. In investigating the problem, I also realized that the way that protobuf packages the protos leads to a circular dependency, so I have moved one proto from `keystore_api` to `message_contents` to resolve the circular dep.

Since the `keystore_api` package is only used in `xmtp-js` beta branch, it will be easy to update to the moved proto.


nakajima (approved):
Cool, makes sense. Thanks for taking a look and getting a fix in so quickly!

PR #72, opened 2023-03-30
Introduce `ephemeral` field to Envelope

nakajima:
This PR proposes adding a new field called `ephemeral` to `Envelope`. In some cases we might want to publish envelopes to the network that don't want to be persisted. Some examples might be typing notifications or presence messages.

Envelopes with `ephemeral` set to true would be published to clients currently subscribed to the topic but would not be written to the store. Queries would not return these envelopes.


nakajima:
PR for node implementation here: https://github.com/xmtp/xmtp-node-go/pull/264

proto/message_api/v1/message_api.proto
```diff
@@ -91,6 +91,10 @@ message Envelope {
     // it MUST be equivalent to the timestamp in the envelope.
     uint64 timestamp_ns = 2;
     bytes message = 3;
+
+    // Ephemeral messages are published to subscribed clients
+    // but are not persisted to the network.
+    optional bool ephemeral = 4;
```
nakajima, line 97: This is the actual change.

nakajima:
> Are all messages permitted to be ephemeral, or only specific message types?

Sorry, I should have closed this. I ended up going a different route and implementing ephemeral topics instead. Here's the JS SDK PR: https://github.com/xmtp/xmtp-js/pull/321.

PR #74, opened 2023-04-03
Add podspec

nakajima:


michaelx11:
What does this do? Does it specify a container + some manifest for Cocoapods generation of Swift code from the protos? Does it have to be top-level?

PR #83, opened 2023-04-27
Protect against changes on beta branch

neekolas:
## Summary

Since v8, `xmtp-js` has been using the `beta` branch and has support for some of the new types in it.

Given the change in direction for the messaging team, and the focus on Signal work, we are not planning on making use of the new `SendKey` types. But given that code expecting those types is already in the wild, we need to reserve the fields to ensure that no future developer will accidentally re-use them (which would cause decoding to fail for v8 clients).

This PR also includes an export that was only present in the `beta` branch, which is required to point `xmtp-js` back at `main`.


michaelx11 (approved):
Excellent heads up awareness here

PR #84, opened 2023-05-10
Generate message api swagger/openapi spec and include in go package

snormore:
This PR updates `dev/go/generate` to also generate a swagger/OpenAPI spec for the `message_api/v1` protos, and embed in the Go package so it can be served along with the API.

Screenshot 2023-05-10 at 10 05 55 AM


proto/buf.yaml
```diff
@@ -2,6 +2,7 @@ version: v1
 name: buf.build/xmtp/proto
 deps:
   - buf.build/googleapis/googleapis
+  - buf.build/grpc-ecosystem/gateway
```
snormore, line 5: Not sure this is right, the GH workflow is 🔴

saulmc:
@snormore @richardhuaaa Is this hosted anywhere that it can be shared with partners who need an HTTP API?

PR #85, opened 2023-05-22
feat: initial group messaging updates

nakajima:
This PR adds the following:

- Ability to create an invite for multiple recipients with the same topic/key
- Ability to create an invite for a recipient for an existing topic/key
- ~~Optional `group_context` field for `InvitationContext` to allow specification of initial group members.~~

~~By adding the `optional` field I ended up needing to add `--experimental_allow_proto3_optional` to the go generation, might need to do that for other languages as well.~~


proto/message_contents/invitation.proto
```diff
@@ -19,13 +19,19 @@ message InvitationV1 {
...
+
     // The context type
     message Context {
         // Expected to be a URI (ie xmtp.org/convo1)
         string conversation_id = 1;
         // Key value map of additional metadata that would be exposed to
         // application developers and could be used for filtering
         map<string, string> metadata = 2;
+
+        optional GroupContext group_context = 3;
```
neekolas, line 34: Do we actually need the `optional` flag here? Objects are always optional in `proto`, and should default to an empty object when read with an unset value. The `optional` flag is really made for scalar fields, where it creates a wrapper object to differentiate between the default value and the value being unset.

The Typescript types do expect that you set a value for an object field, but we can work around that at the SDK layer and set our own defaults

proto/message_contents/invitation.proto
```diff
@@ -19,13 +19,19 @@ message InvitationV1 {
         bytes key_material = 1;  // randomly generated key material (32 bytes)
     }
 
+    message GroupContext {
```
neekolas, line 22: Is there a reason we can't just do this with the existing `metadata` field? While, the value is required to be a string, it doesn't seem like the most onerous thing in the world to just have it be a comma separated list

proto/keystore_api/v1/keystore.proto
```diff
@@ -116,12 +116,32 @@ message CreateInviteRequest {
     uint64 created_ns = 3;
 }
 
+// Request to create an invite payload from an existing topic, and store the topic keys in the Keystore
+message CreateInviteFromTopicRequest {
+    xmtp.message_contents.InvitationV1.Context context = 1;
```
neekolas, line 121: Do we actually need the context here? The Keystore already should already have the `context` in its records, if it knows the `content_topic`. Same for `created_ns`.

nakajima: Ya, I think you're right. I can remove these.

proto/message_contents/invitation.proto
nakajima: Gotcha, I'm not what you'd call a protobuf expert so I appreciate that context. I can remove the optional (and the flag passing in the go generation).

proto/message_contents/invitation.proto
nakajima: I considered metadata for this, but it felt a bit too loose to me for such a core concept as group membership. I could be convinced otherwise but to me it seemed like a good idea to have something concrete in the protocol to determine if something is a group or not.

dmccartney, line 22: tl/dr: I would avoid the word "Group" here entirely.
The "group" metaphor (to me at least) implies a set of "members" who can be listed/added/removed etc. But if I understand correctly, this won't support listing or removing members. Having a field called "GroupContext" seems like a footgun ("oh cool, XMTP supports group messaging, how do I build the group list? ah! `GroupContext` I'll use that to show the member list").  "Conversation" already implies (to me at least) the possibility of multiple participants.

nakajima:
Updated to remove the GroupContext as well as the redundant fields in CreateInviteFromTopicRequest.

proto/keystore_api/v1/keystore.proto
```diff
@@ -116,12 +116,30 @@ message CreateInviteRequest {
     uint64 created_ns = 3;
 }
 
+// Request to create an invite payload from an existing topic, and store the topic keys in the Keystore
+message CreateInviteFromTopicRequest {
+    xmtp.message_contents.SignedPublicKeyBundle recipient = 2;
```
neekolas, line 121: We can reset the numbering here so that `recipient` is in field 1

PR #86, opened 2023-05-31
feat: add signature

neekolas:
## Summary

- Adds EIP-191 association type for wallet signatures
- Renames contact bundle to `InstallationContactBundle`
- Wraps the top level contact in a `version` union to allow for upgradeability later. Given that this is a top-level proto (it gets decoded directly), we need a way to make breaking changes that doesn't involve rotating the topic
- Fix go package versioning.

### DIFF
--- a/proto/v3/message_contents/association.proto
+++ b/proto/v3/message_contents/association.proto
@@ -1,0 +1,20 @@
+// Association types
+syntax = "proto3";
+
+package xmtp.v3.message_contents;
+
+option go_package = "github.com/xmtp/proto/v3/go/v3/message_contents";
+option java_package = "org.xmtp.proto.v3.message.contents";
+
+// EIP191Association is used for all EIP 191 compliant wallet signatures
+message Eip191Association {
+    // The text that was signed, not including the prefix '\0x19Ethereum....'
+    string text = 1;
+    RecoverableEcdaSignature signature = 2;
+}
+
+// RecoverableEcdsaSignature
+message RecoverableEcdaSignature {
+    // Includes recovery id as the last byte
+    bytes bytes = 1;
+}
--- a/proto/v3/message_contents/public_key.proto
+++ b/proto/v3/message_contents/public_key.proto
@@ -4,8 +4,9 @@ syntax = "proto3";
 
 package xmtp.v3.message_contents;
 
-// v4 here since it was already v3 for golang prior to this
-option go_package = "github.com/xmtp/proto/v4/go/message_contents";
+import "v3/message_contents/association.proto";
+
+option go_package = "github.com/xmtp/proto/v3/go/v3/message_contents";
 option java_package = "org.xmtp.proto.v3.message.contents";
 
 // An unsigned public key used by libxmtp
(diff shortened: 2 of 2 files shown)


proto/v3/message_contents/public_key.proto
```diff
@@ -4,8 +4,9 @@ syntax = "proto3";
 
 package xmtp.v3.message_contents;
 
-// v4 here since it was already v3 for golang prior to this
-option go_package = "github.com/xmtp/proto/v4/go/message_contents";
+import "v3/message_contents/association.proto";
+
+option go_package = "github.com/xmtp/proto/v3/go/v3/message_contents";
```
neekolas, line 9: The `go_package` field is sensitive. It can't be `github.com/xmtp/proto/v4` unless we increment the version on the git tag to a new major version. Until then, needs to stay as `v3`.

proto/v3/message_contents/association.proto
```diff
@@ -0,0 +1,26 @@
...
+// Allows for us to update the format of the association text without
+// incrementing the entire proto
+enum AssociationTextVersion {
+    ASSOCIATION_TEXT_VERSION_UNSPECIFIED = 0;
+    ASSOCIATION_TEXT_VERSION_1 = 1;
+}
+
+// EIP191Association is used for all EIP 191 compliant wallet signatures
+message Eip191Association {
+    AssociationTextVersion association_text_version = 1;
```
neekolas, line 18: I'm only storing the `version` of the text and not the actual text signed. This should be enough information to reconstruct the text in the recipient (they already know the public key bytes and address from other fields in the message). Any differences in output will lead to an error verifying the signature.

proto/v3/message_contents/public_key.proto
```diff
@@ -59,8 +61,15 @@ message VmacFallbackKeyRotation {
 }
 
 // A contact bundle served by the server to a requesting client
-message VmacContactBundle {
+message VmacInstallationPublicKeyBundleV1 {
     VmacAccountLinkedKey identity_key = 1;
-    // An unused one time prekey or fallback key returned by the server
-    VmacDeviceLinkedKey prekey = 2;
+    VmacDeviceLinkedKey fallback_key = 2;
```
neekolas, line 66: We've been previously using the account's `fallback_key` and calling it the `prekey` in the proto, so this renames for clarity.

Also opens the door for us to add actual `prekey`s into the message later if we need.

neekolas:
> Could we go ahead and rename VmacDeviceLinkedKey to VmacInstallationLinkedKey as well?

@richardhuaaa done

proto/v3/message_contents/association.proto
```diff
@@ -0,0 +1,26 @@
...
+
+// Allows for us to update the format of the association text without
+// incrementing the entire proto
+enum AssociationTextVersion {
+    ASSOCIATION_TEXT_VERSION_UNSPECIFIED = 0;
+    ASSOCIATION_TEXT_VERSION_1 = 1;
+}
+
+// EIP191Association is used for all EIP 191 compliant wallet signatures
+message Eip191Association {
```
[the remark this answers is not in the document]
neekolas: I've updated to have a `oneof` in the `VmacAccountLinkedKey`

proto/v3/message_contents/association.proto
jazzz: I think this could work for the Static Signature type in the short term, but this structure would limit the flexibility of the Association system. Namely in that there is an assumption of the association text which breaks the push towards self-describing associations. 

I get the push towards a functional E2E message flow so this makes sense today, but I think we'll need to take another look prior to release.

neekolas: That's true. I've been deliberately leaving other signature types out of my thinking, since I'm not sure which ones we'll need and when. 

I did make a oneof so we can add more concrete types as needed.

PR #89, opened 2023-06-14
Wallet address on association

neekolas:
## Summary

In order to make `InstallationContactBundle` fully self-describing and able to be verified, we need to include the `wallet_address` on the association.

Previously we were able to get the intended wallet address for an association from context (for example, the topic the message was found on), but in the case of Invite messages we need it to be explicit.

In either case, if the wallet address does not match the signature, verification will fail and the contact will be rejected.

## Alternatives considered
We _could_ store the full text that was signed instead. This is even longer (containing both the wallet address and public key), and it's messier since we would have to parse and validate that the text conforms to our standards for association text. 

Just storing the hash that was signed is not an option because we need to verify that the signature is coming from a well-formed association text referring to the right wallet and public key.

### DIFF
--- a/proto/v3/message_contents/association.proto
+++ b/proto/v3/message_contents/association.proto
@@ -17,6 +17,7 @@ enum AssociationTextVersion {
 message Eip191Association {
     AssociationTextVersion association_text_version = 1;
     RecoverableEcdsaSignature signature = 2;
+    string wallet_address = 3;
 }
 
 // RecoverableEcdsaSignature


proto/v3/message_contents/association.proto
```diff
@@ -17,6 +17,7 @@ enum AssociationTextVersion {
 message Eip191Association {
     AssociationTextVersion association_text_version = 1;
     RecoverableEcdsaSignature signature = 2;
+    string wallet_address = 3;
```
jazzz, line 20: Originally the intention was that the AssociationText would be self describing. It contains the linkage between the two entities and the signature provides the proof: https://github.com/xmtp/libxmtp/blob/main/xmtp/src/association.rs#L112-L117

> it's messier since we would have to parse and validate that the text conforms to our standards for association text.

Worth noting that the Association design is based on this principal.  Each implementation of the AssociationText has the logic to validate that the text conforms to its standard. This abstraction was put into place to support Eip1271, SIW*, multichain and other signing standards as needed. Validating expiration dates is a good example of something the AssociationText would need to validate.

As context each Association was intended to contain 6 pieces of data
- WalletAddr
- KeyBytes
- SigningTextGeneration Procedure
- SigningTextValidation Procedure
- SignatureBytes
- Signature Validation Procedure

For StaticSignatures you are right the text can be generated so I agree theres no need to replicate it. However including the keyBytes and the wallet address would make the Associations self-describing. How the information is organized/structured is largely preference, and I don't think i'm too picky.

jazzz (approved):
No real blockers, however it seems we are circling the drain on the Association concept. 
- If it doesn't fit our needs then lets get rid of it; 
- If it does then leaning into it would achieve the full benefits

Currently right now it seems like we are stuck in the middle.

proto/v3/message_contents/association.proto
neekolas: > However including the keyBytes and the wallet address would make the Associations self-describing.

Ya, I guess it's just a question of whether we want to duplicate the key bytes into the association. With this PR the `VmacAccountLinkedKey` has all the information to verify itself...we just have to pick and choose data from different parts of it. It's a little less pure, but it shaves some bytes off of every contact. And we don't have to worry about cases where the pub_key in the signature text doesn't match the pub key in the parent message.

I'm alright with leaving this as-is for the simple case of static EIP-191 signatures. For more advanced signature types like SIWE, we can choose to store more data if it makes sense. The code is structured nicely so that each association type can have its own data structures and verification logic.

PR #92, opened 2023-07-26
feat: add padlock message types

richardhuaaa:
Adds new message protos for Padlock messages

### DIFF
--- a/proto/v3/message_contents/association.proto
+++ b/proto/v3/message_contents/association.proto
@@ -28,5 +28,5 @@ message RecoverableEcdsaSignature {
 
 // EdDSA signature bytes matching RFC 8032
 message EdDsaSignature {
-    bytes byte = 1;
+    bytes bytes = 1;
 }


proto/message_contents/message.proto
```diff
@@ -67,4 +67,38 @@ message DecodedMessage {
...
+    // sender of the message.
+    string session_id = 2;
+}
+
+// Encrypted body included with messages, only visible to recipients
+// - For safety, the sender_account_id and sender_installation_id MUST be
+//   derived from the session_id in the header by each installation receiving
+//   the message
+// - The recipient is unnecessary to include here - the recipient at the session
+//   level is the receiving installation, and the recipient at the conversation
```
jazzz, line 85: "the recipient at the conversation level is whoever is the conversation refered to by the convo_id" 
Is this true?
I would have assumed that the recipient is the account(s) which is associated by the recieving installation.

proto/message_contents/message.proto
```diff
@@ -67,4 +67,38 @@ message DecodedMessage {
...
+//   level is whoever is in the conversation referred to by convo_id (multiple
+//   people in the case of groups)
+// - The recipient installation MUST cross-check that the sender_account_id is a
+//   valid participant of the conversation referred to by convo_id
+// - The recipient installation MUST validate that the MessageHeader hashes to
+//   the header_hash.
+//   TODO is that sufficient to replace the AssociatedData in v2?
```
jazzz, line 92: Is there a reason to use the hash over the data itself. Supplying the data as associated data authenticates that it was not tampered with. Using a hash in this case adds complexity. I'd say supplying the header itself would be a more streamlined solution.

proto/message_contents/message.proto
```diff
@@ -67,4 +67,38 @@ message DecodedMessage {
...
+
+// Plaintext header included with messages, visible to all
+message PadlockMessageHeader {
+    uint64 sent_ns = 1;
+    // The session_id is derived via SHA256. Any other identifier will leak the
+    // sender of the message.
+    string session_id = 2;
+}
+
+// Encrypted body included with messages, only visible to recipients
```
jazzz, line 80: How is this message encrypted? Using an AEAD scheme?

proto/message_contents/message.proto
```diff
@@ -67,4 +67,38 @@ message DecodedMessage {
...
+//   level is whoever is in the conversation referred to by convo_id (multiple
+//   people in the case of groups)
+// - The recipient installation MUST cross-check that the sender_account_id is a
+//   valid participant of the conversation referred to by convo_id
+// - The recipient installation MUST validate that the MessageHeader hashes to
+//   the header_hash.
+//   TODO is that sufficient to replace the AssociatedData in v2?
+message PadlockMessagePayload {
+    uint32 message_version = 1;
+    bytes header_hash = 2;      // PadlockMessageHeader
```
jazzz, line 95: From what I gather here, the intention is to move away from an AEAD encryption scheme for messages and as such a mechanism for authenticating the header is needed. Did I understand that correctly?

And so the question is, "is a hash of the header payload sufficient for authenticating its contents?

 I'd want to see the hash value contexted to minimize the threat of hash reuse.

proto/message_contents/message.proto
```diff
@@ -67,4 +67,38 @@ message DecodedMessage {
...
     ConversationReference conversation = 7;
     bytes content_bytes = 8;  // encapsulates EncodedContent
+}
+
+// Plaintext header included with messages, visible to all
+message PadlockMessageHeader {
+    uint64 sent_ns = 1;
+    // The session_id is derived via SHA256. Any other identifier will leak the
+    // sender of the message.
+    string session_id = 2;
```
jazzz, line 77: What is the sessionId in this case --  the hash of the installationPublicKey?

Is there a reason why this data is unencrypted?

proto/message_contents/message.proto
```diff
@@ -67,4 +67,38 @@ message DecodedMessage {
...
+//   level is the receiving installation, and the recipient at the conversation
+//   level is whoever is in the conversation referred to by convo_id (multiple
+//   people in the case of groups)
+// - The recipient installation MUST cross-check that the sender_account_id is a
+//   valid participant of the conversation referred to by convo_id
+// - The recipient installation MUST validate that the MessageHeader hashes to
+//   the header_hash.
+//   TODO is that sufficient to replace the AssociatedData in v2?
+message PadlockMessagePayload {
+    uint32 message_version = 1;
```
neekolas, line 94: Seems like we can make this an enum rather than a uint32

proto/message_contents/message.proto
```diff
@@ -67,4 +67,38 @@ message DecodedMessage {
...
+    bytes header_hash = 2;      // PadlockMessageHeader 
+    string convo_id = 3;
+    bytes content_bytes = 4;    // EncodedContent
+}
+
+// Combines the plaintext header with the encrypted payload
+message PadlockMessageEnvelope {
+    bytes header_bytes = 1;     // PadlockMessageHeader
```
neekolas, line 102: If we aren't using the header bytes as AEAD, why not just include the full header as a field?

The reason we had `header_bytes` in V2 was to remove the possibility that two proto implementations would serialize differently, causing authenticated encryption to fail.

proto/message_contents/message.proto
jazzz: That makes sense. I think I'm getting pedantic around conversation membership. Where the recipient is unknown with the exception that if It was sent to you it was intended for you. Any conversationId would simply hint at a possible set of recipients, rather than being a guaranteed.

I don't think it particularly matters in this case, but would have more of an impact in the context of threat and privacy modelling

proto/message_contents/message.proto
jazzz: The issue here is that then the session_id becomes an identifying piece of metadata, and would leak information about pseudonymous conversation activity. Which intern could Identity senders through simple analysis.

- search inbound topics for messages with the same session_id
- Find Association for a given installationId

If the intention is to meet our sender privacy goals then session_id needs to be considered private. If the intention is to address those in a fast follow, then the header could contain more information such as the actual sending user, the conversation_id, etc which would make a sealed sender implementation in the future easier.

proto/message_contents/message.proto
```diff
@@ -67,4 +67,38 @@ message DecodedMessage {
...
+// - The recipient is unnecessary to include here - the recipient at the session
+//   level is the receiving installation, and the recipient at the conversation
+//   level is whoever is in the conversation referred to by convo_id (multiple
+//   people in the case of groups)
+// - The recipient installation MUST cross-check that the sender_account_id is a
+//   valid participant of the conversation referred to by convo_id
+// - The recipient installation MUST validate that the MessageHeader hashes to
+//   the header_hash.
+//   TODO is that sufficient to replace the AssociatedData in v2?
+message PadlockMessagePayload {
```
jazzz, line 93: The message payload is missing strong protection from an unknown keyshare attack. The Message needs to contain sender and receiver identifying information to ensure that the message can be validated.

jazzz: Unless that is occurring at a lower encapsulated layer in VMAC?

proto/message_contents/message.proto
```diff
@@ -67,4 +67,38 @@ message DecodedMessage {
     string content_topic = 6;
     ConversationReference conversation = 7;
     bytes content_bytes = 8;  // encapsulates EncodedContent
+}
+
+// Plaintext header included with messages, visible to all
+message PadlockMessageHeader {
+    uint64 sent_ns = 1;
```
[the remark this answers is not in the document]
neekolas: The Waku envelope definitely has a timestamp. In V2 we enforce it is the same as any timestamps in the header.

proto/message_contents/message.proto
jazzz: Followup. My assumption was incorrect here. VMAC uses AES-CBC with provides confidentiality only, and using AssociatedData to authenticate the headers is not an option.

proto/v3/message_contents/message.proto
```diff
@@ -0,0 +1,61 @@
...
+// 3. Use the session to decrypt the payload
+// 4. Verify that the header_signature in the decrypted payload matches the
+//    header bytes
+// 5. Verify that the signer of the header_signature matches the
+//    sender_installation_id
+// 6. Verify that both the sender_user and recipient_user are partipants of the
+//    conversation referenced by convo_id
+message PadlockMessagePayload {
+    PadlockMessagePayloadVersion message_version = 1;
+    RecoverableEcdsaSignature header_signature = 2; // Sign PadlockMessageHeader
```
jazzz, line 52: I think this signature will need some extra thought here. Its not immediately clear to me that we want to use ECDSA here. V3 Installations use ed25519, so using ECDSA would imply that padlock messages are signed by the V2Identity (or a derived key)

proto/message_contents/message.proto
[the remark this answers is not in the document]
jazzz: While I am a huge proponent of completely isolating the messaging protocol from implementation details of the delivery service. I'm curious how the current concept of invitations matches with this. 

My understanding was that invitations were required to arrive on a special topic, which would mean that the messages and waku envelopes are dependent on one another. 


Is there flexibility to change the protobuf later? or does it need to be codified now?

jazzz (approved):

PR #94, opened 2023-08-22
Add new Subscribe2 method

neekolas:
## Summary

Our current Subscribe API is limited by the fact that it was built for browser clients. You cannot update subscriptions. To add or remove topics from a subscription, you have to close the existing subscription and create a new one. This is because browsers do not support bidirectional GRPC streaming. With iOS, Android, and Rust native supporting full-fledged GRPC, this limitation doesn't apply.

This PR creates a new SubscribeV2 GRPC service that allows clients to update their subscriptions without disconnecting. These small disconnects are a pain point for apps who rely on streams to get new messages. When using `conversations.streamAllMessages` these disconnects happen every time a new conversation is started, since the stream needs to include the new conversation topics.

This change has the added benefit of decreasing load on our nodes, since we wouldn't have to tear down and recreate the subscriptions every time a topic is added.


proto/message_api/v1/message_api.proto
```diff
@@ -7,139 +7,139 @@ import "protoc-gen-openapiv2/options/annotations.proto";
 
 option go_package = "github.com/xmtp/proto/v3/go/message_api/v1";
 option java_package = "org.xmtp.proto.message.api.v1";
-
 option (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_swagger) = {
-    info: {
-        title: "MessageApi";
-        version: "1.0";
-    };
+  info: {
```
neekolas, line 11: Not sure why my formatter is different from what was used for this file. We should probably add some vscode config to standardize everyone

proto/message_api/v1/message_api.proto
```diff
@@ -7,50 +7,54 @@ import "protoc-gen-openapiv2/options/annotations.proto";
...
-        body: "*"
-    };
+            post: "/message/v1/subscribe"
+            body: "*"
+        };
     }
+    // Subscribe to a stream of new envelopes and your subscription using
+    // bidirectional streaming
+    // protolint:disable:next RPC_REQUEST_STANDARD_NAME
+    rpc SubscribeV2(stream SubscribeRequest) returns (stream Envelope) {}
```
mkobetic, line 37: Nitpick: How about `Subscribe2`, V2 makes it sound like there potentially will be V3 at some point, also might be confusing with relation to protocol versions.

proto/message_api/v1/message_api.proto
```diff
@@ -7,50 +7,54 @@ import "protoc-gen-openapiv2/options/annotations.proto";
...
-        title: "MessageApi";
-        version: "1.0";
-    };
-};
+    info: { title: "MessageApi";
+version: "1.0";
+}
+;
+}
+;
```
mkobetic, line 16: Is this what the formatter is forcing? This looks terrible.

mkobetic (approved):

proto/message_api/v1/message_api.proto
neekolas: Works for me

proto/message_api/v1/message_api.proto
neekolas: Completely agree. I tried using a different formatter and that one fails the linter. So, I'm kinda stuck.

PR #97, opened 2023-09-15
Add legacy V2 identity key association

richardhuaaa:
Needed to test out v3 in current SDK's without introducing breaking changes. Publishing now to avoid conflict with Daria's ramp-up task.

### DIFF
--- a/proto/v3/message_contents/association.proto
+++ b/proto/v3/message_contents/association.proto
@@ -13,6 +13,7 @@ option java_package = "org.xmtp.proto.v3.message.contents";
 enum AssociationTextVersion {
     ASSOCIATION_TEXT_VERSION_UNSPECIFIED = 0;
     ASSOCIATION_TEXT_VERSION_1 = 1;
+    ASSOCIATION_TEXT_VERSION_LEGACY = 100;
 }
 
 // EIP191Association is used for all EIP 191 compliant wallet signatures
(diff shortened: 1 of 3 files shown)


proto/v3/message_contents/association.proto
```diff
@@ -6,18 +6,35 @@ package xmtp.v3.message_contents;
...
-    string wallet_address = 3;
+    oneof association_data {
+        CreateIdentityAssociationData create_identity_data = 3;
+        InstallationGrantAssociationData installation_grant_data = 4;
+    }
+
+    // Legacy - all of the information needed to construct the Create Identity
+    // association text used in XMTP v2
+    message CreateIdentityAssociationData {
+        string wallet_address = 1;
```
neekolas, line 21: This is confusing to me, since the Create Identity association text is actually the encoded protobuf of the public key, not the wallet address.

proto/v3/message_contents/association.proto
```diff
@@ -6,18 +6,35 @@ package xmtp.v3.message_contents;
...
-// incrementing the entire proto
-enum AssociationTextVersion {
-    ASSOCIATION_TEXT_VERSION_UNSPECIFIED = 0;
-    ASSOCIATION_TEXT_VERSION_1 = 1;
-}
-
 // EIP191Association is used for all EIP 191 compliant wallet signatures
 message Eip191Association {
-    AssociationTextVersion association_text_version = 1;
+    reserved 1;
```
neekolas, line 11: Do we care about reserving fields if we aren't yet worried about interoperability between existing versions. Think we can just throw away all our existing v3 keys and start fresh

PR #99, opened 2023-10-02
Added proto

daria-github:
Added proto to close https://github.com/xmtp/libxmtp/issues/257

### DIFF
--- a/proto/v3/message_contents/public_key.proto
+++ b/proto/v3/message_contents/public_key.proto
@@ -34,6 +34,17 @@ message VmacAccountLinkedKey {
     }
 }
 
+// Keys linked to an account (e.g. signed by a wallet)
+// The purposes of the keys are encoded in the signatures
+message VmacAccountLinkedKeys {
+    VmacUnsignedPublicKey curve25519_key = 1;
+    VmacUnsignedPublicKey ed25519_key = 2;
+    
+    oneof association {
+        Eip191Association eip_191 = 3;
+    }
+}
+
 // A key linked to an installation (e.g. signed by an installation identity key)
 // The purpose of the key is encoded in the signature
 message VmacInstallationLinkedKey {


proto/v3/message_contents/public_key.proto
```diff
@@ -34,6 +34,17 @@ message VmacAccountLinkedKey {
     }
 }
 
+// Keys linked to an account (e.g. signed by a wallet)
+// The purposes of the keys are encoded in the signatures
+message VmacAccountLinkedKeys {
```
neekolas, line 39: This creates a new type for the keys, but I don't see where this will include it in the contact bundle.

proto/v3/message_contents/public_key.proto
```diff
@@ -34,6 +34,17 @@ message VmacAccountLinkedKey {
...
+    VmacUnsignedPublicKey curve25519_key = 1;
+    VmacUnsignedPublicKey ed25519_key = 2;
```
neekolas, line 41: A `VmacUnsignedPublicKey` only has `curve25519` as an option for the key bytes, but doesn't have the option for `ed25519`. The field is just `bytes`, so we can shove anything we want in there, but we should allow it to mark the keys as being of different types in that message to avoid confusion. Otherwise you are going to access `key.ed25519_key.curve25519`, which is misleading and confusing.

proto/v3/message_contents/public_key.proto
richardhuaaa: It might make sense to copy `VmacInstallationPublicKeyBundleV1` to a `V2` struct and just replace the `identity_key` with something like `installation_keys`. The V1 struct can just be deleted later

neekolas: Given we are offering no guarantees of compatibility right now, I'd also be OK with just replacing V1 with the new type.

neekolas: Keeping old versions around isn't all that helpful if we don't plan to have a SDK available that can read them

richardhuaaa: I thought of that too, but with that approach as soon as we call `gen_protos.sh` we have to immediately refactor to use the new type in all locations in the codebase. With the separate type, we can replace V1 with V2 in pieces in the code until there's no V1 left, and then delete the V1 proto, so we don't have any baggage.

neekolas: We aren't going to have to update in one pass regardless? Seems like things will be pretty broken with half the contact handling code on V1 and half on V2. 

If we introduce a v2 and then delete v1, there would be an extra step of renaming v2 to v1 at the end. Or we just launch with only a v2 proto and a reserved v1 slot lives there forever.

richardhuaaa:
This is no longer relevant after shifting to MLS - thanks for putting this together!

PR #100, opened 2023-10-04
Signed ECIES Payloads

neekolas:
## Summary

- Adds a new `SignedPayload` message type to hold arbitrary bytes and a signature
- Adds a wrapper type for `EciesMessage`, which currently just holds a `v1` structure for encrypted `SignedPayload` messages

### DIFF
--- a/.github/workflows/buf.yml
+++ b/.github/workflows/buf.yml
@@ -12,6 +12,9 @@ jobs:
       - uses: actions/checkout@v2
       # Install the `buf` CLI
       - uses: bufbuild/buf-setup-action@v1
+        with:
+          github_token: ${{ github.token }}
+          buf_token: ${{ secrets.BUF_TOKEN }}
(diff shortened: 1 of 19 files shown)


nakajima (approved):

richardhuaaa:
@neekolas this is for the v2 portable preferences right? Does this belong in `main`?

neekolas:
You would think so, but libxmtp only looks at the v3 branch. Including in libxmtp's V2 branch. 

We could change that, since it's quite confusing.

richardhuaaa:
Ahhh gotcha. Fine with leaving it as-is for now

neekolas:
Really what I'd like to do is kill the v3 proto branch. None of our v3 work to date has involved changing any v2 protos, and I don't think it will. So the separate branching strategy isn't protecting us from much.

PR #103, opened 2023-10-13
Update proto messages

neekolas:
## Summary

Updates the Protobuf message types in V3.

The goals of these changes are:

1. Ensure backwards/forwards compatibility for outermost message envelopes. We only have a single inbound topic in V3, so we want to avoid protobuf decoding failures for any valid message sent to that topic.
2. Add support for message types that live outside of a conversation. For example, `MessageBackupRequest` and `MessageBackupResponse` messages are always sent from another installation and don't fit into the typical conversation model. They don't need to be wrapped in `EncodedContent`
3. Add support for [SealedSender](https://github.com/xmtp/libxmtp/pull/266) on headers
4. Remove any mention of Padlock

## Notes
I copied the `EncodedContent` message format in to the V3 folder verbatim. This is mostly just so that V3 never has to reference V2 types, which allows us to import more narrowly and will make the eventual deprecation of V2 types easier.

### DIFF
--- a/proto/v3/message_contents/message.proto
+++ b/proto/v3/message_contents/message.proto
@@ -60,19 +60,20 @@ message MessagePayload {
   EdDsaSignature header_signature = 1; // Signed MessageHeader
   oneof contents {
     ConversationMessageV1 conversation_message_v1 = 2;
-    // TODO: Add other message types (Backup Request, P4, Backup Response)
+    // TODO: Add other message types
+    // (Message Backup Request, P4, Message Backup Response)
   }
 }
 
 // Combines the plaintext header with the encrypted payload
 message MessageEnvelope {
   // First version of the envelope format
-  message V1Envelope {
+  message EnvelopeV1 {
     bytes header_bytes = 1; // MessageHeader
     bytes ciphertext = 2; // Encrypted MessagePayload
   }
 
   oneof envelope {
-    V1Envelope v1 = 1;
+    EnvelopeV1 v1 = 1;
   }
 }
--- a/proto/v3/message_contents/message.proto
+++ b/proto/v3/message_contents/message.proto
@@ -12,7 +12,7 @@ option java_package = "org.xmtp.proto.v3.message.contents";
 // Plaintext header included with messages, visible to all
 // Recipients can verify this header has not been tampered with.
 // Servers are unable to verify if the header has been tampered with.
-message MessageHeader {
+message MessageHeaderV1 {
   // SealedSender wrapper for arbitrary bytes
   message SealedSender {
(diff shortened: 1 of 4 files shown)


nplasterer (approved):

proto/v3/message_contents/message.proto
```diff
@@ -4,33 +4,42 @@ syntax = "proto3";
...
-
 // Plaintext header included with messages, visible to all
 // Recipients can verify this header has not been tampered with.
 // Servers are unable to verify if the header has been tampered with.
-message PadlockMessageHeader {
+message MessageHeaderV1 {
+  // SealedSender wrapper for arbitrary bytes
+  message SealedSender {
+    bytes ephemeral_public_key = 1;
+    bytes ephemeral_static_key = 2;
```
richardhuaaa, line 19: What is the `ephemeral_static_key`?

proto/v3/message_contents/message.proto
```diff
@@ -4,33 +4,42 @@ syntax = "proto3";
...
+
+  // Metadata that is encrypted via SealedSender and only visible to
+  // the recipient
+  message Metadata {
+    string sender_user_address = 1;
+    string sender_installation_id = 2;
+    string recipient_user_address = 3;
+    string recipient_installation_id = 4;
+    bool is_prekey_message = 5;
+  }
```
richardhuaaa, line 31: Are we feeding this whole blob into AES as the equivalent of the `senderIdentityPublic` input in https://signal.org/blog/sealed-sender/, or as the equivalent of the `sender_certificate`?

proto/v3/message_contents/message.proto
```diff
@@ -47,15 +56,26 @@ enum PadlockMessagePayloadVersion {
 //    sender_installation_id
 // 5. Verify that both the sender_user and recipient_user are partipants of the
 //    conversation referenced by convo_id
-message PadlockMessagePayload {
-  PadlockMessagePayloadVersion message_version = 1;
-  EdDsaSignature header_signature = 2; // Signs PadlockMessageHeader
-  string convo_id = 3;
-  bytes content_bytes = 4; // EncodedContent
+message MessagePayloadV1 {
+  EdDsaSignature header_signature = 1; // Signature of the header_bytes
```
richardhuaaa, line 60: Depending on the details of how we are implementing sealedsender, I'm wondering if we still need to sign this

richardhuaaa (commented):
Like these changes overall! Thanks for the continuous cleanup you've been doing

neekolas:
Closing this to focus on MLS work

PR #106, opened 2023-10-19
Add MLS Service

neekolas:
## Summary

- Adds a MLS delivery service, with a moderately thought through API spec. Subject to change, but I want to get something in to scaffold the service, even if some of the endpoints wind up changing.
- Adds a few wrapper types for the message formats that will be published to the network


.protolint.yaml
```diff
@@ -49,7 +50,7 @@ lint:
     # MAX_LINE_LENGTH rule option.
     max_line_length:
       # Enforces a maximum line length
-      max_chars: 80
+      max_chars: 150
```
neekolas, line 53: Realized that the new Protobuf formatter doesn't necessarily split lines (for example, for long RPC definitions)

proto/message_api/v3/mls.proto
```diff
@@ -0,0 +1,161 @@
+// Message API
+syntax = "proto3";
+package xmtp.message_api.v3;
```
neekolas, line 3: I just skipped right over V2 and went straight to V3 for naming consistency with other V3 protos

proto/message_api/v3/mls.proto
```diff
@@ -0,0 +1,180 @@
...
+    title: "MlsApi";
+    version: "1.0";
+  };
+};
+
+// RPCs for the new MLS API
+service MlsApi {
+  // Publish a MLS payload, that would be validated before being stored to the
+  // network
+  rpc Publish(PublishRequest) returns (google.protobuf.Empty) {
```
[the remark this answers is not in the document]
neekolas: I went with `PublishToGroup`, since I realized we are going to need a different endpoint for Welcome messages.

proto/mls/message_contents/message.proto
```diff
@@ -0,0 +1,32 @@
+// V3 invite message structure
+syntax = "proto3";
+
+package xmtp.mls.message_contents;
+
+option go_package = "github.com/xmtp/proto/v3/go/mls/message_contents";
+option java_package = "org.xmtp.proto.mls.message.contents";
+
+// Welcome messages are sent to an installation's welcome topic and are
+// encrypted using Noise Pattern N
```
richardhuaaa, line 10: What metadata are we trying to hide here? Curious why we need to add this on top of MLS

richardhuaaa (approved):
Happy to get something in and iterate.

One idea I'd like to plant here is if we are able to make any of these endpoints less MLS-specific? For example, by introducing a generic OrderedTopic concept, or a generic SingleUseTopic?

proto/mls/message_contents/message.proto
neekolas: Maybe I was mistaken. I thought there was some sensitive metadata unencrypted in the Welcome. But looking a the code, it's [all encrypted](https://github.com/openmls/openmls/blob/2d4991b53feb3f8490ca23fbb5a5ccfff094d07d/openmls/src/messages/mod.rs#L66). The only thing they really expose is the `KeyPackageRef` of the recipient, and the ciphersuites used. That's not sensitive at all.

neekolas: Will remove references to Noise for now and leave the Welcomes unencrypted. We may have to re-add if we add our own metadata around the Welcome for some reason (not sure if we have to validate the sender with a signature to know who invited you)

neekolas:
> One idea I'd like to plant here is if we are able to make any of these endpoints less MLS-specific? For example, by introducing a generic OrderedTopic concept, or a generic SingleUseTopic?

I'm open to ordered topics. I'd like to get a better understanding of what you'd want a `SingleUseTopic` to do.

Even with that, I think the `PublishToGroup` is going to be pretty MLS-specific until we have a better commit resolution strategy. That one requires not just ordering, but also rejection of messages from past epochs so that the client can refresh their state and retry the request.

richardhuaaa:
Just spitballing, a `SingleUseTopic` deletes payloads as they are read. An `OrderedTopic` requires payloads to have two fields, `currentEpoch` and `nextEpoch`, and publishes fail if `currentEpoch` doesn't match the current epoch. I say this with low confidence, I think you're deeper into what the actual validation logic would be than I am

PR #107, opened 2023-10-19
MLS Protos

neekolas:
## Summary

- Protobuf API for the Rust service that will be consumed by our Nodes.
- Updates the schema for `RegisterInstallation` to accept a last resort key package, instead of the credential/public key
- Adds new MLS identity protos

See discussion in https://github.com/xmtp/xmtp-node-go/issues/310


proto/mls_validation/v1/service.proto
```diff
@@ -0,0 +1,62 @@
...
+}
+
+// Response to ValidateGroupMessagesRequest
+message ValidateGroupMessagesResponse {
+  // An individual response to one message
+  message ValidationResponse {
+    bool is_ok = 1;
+    string error_message = 2;
+    string group_id = 3;
+    uint64 epoch = 4;
```
richardhuaaa, line 58: Same question here - is this the current expected epoch, or the potential next epoch after a commit is applied?

neekolas: It's whatever was in the message itself, which I believe is the current epoch but should probably confirm.

neekolas: Is the current epoch btw

richardhuaaa (approved):

PR #108, opened 2023-10-24
fix: add the appropriate files so that they get generated

nplasterer:
I wasn't seeing the new proto files in the latest kotlin release because we didn't tell them to generate. This should fix that.


nplasterer:
We commit the swift stuff in this repo so you can use spm to access them.

nakajima (approved):

PR #112, opened 2023-10-25
build: publish dart on release, fix semver for kt and dart

dmccartney:
This fixes two issues:
- `publish-dart.yml` wasn't being triggered because it was waiting for a new `v#.#.#` tag which was being pushed by the `release.yml` workflow. But GH actions generally can't trigger other GH actions (to avoid infinite robot dance loops)
- the kotlin build was publishing artifacts with the prior version because it wasn't waiting for the JS build to first calculate the new `v`

This fixes both by moving dart publishing into the `release.yml` workflow and sequencing both the `dart` and `kotlin` jobs to run after the `js` job publishes the new `v#.#.#` tag.


nplasterer (approved):
Feel like I learn a new GitHub Actions power tool every day!

dmccartney:
Blarg, this failed (`pub.dev` rejects GH actions that aren't directly triggered by a `tag` push -- so it can't come from this workflow. I'm going to manually publish for now and revisit this later.

PR #117, opened 2023-12-07
Make Add/Remove member intents store wallet address

insipx:
Merges into Nick's branch which the protos are generated from in `xmtp_mls` currently.

Companion to https://github.com/xmtp/libxmtp/pull/363

### DIFF
--- a/proto/mls/database/intents.proto
+++ b/proto/mls/database/intents.proto
@@ -22,7 +22,7 @@ message SendMessageData {
 message AddMembersData {
   // V1 of AddMembersPublishData
   message V1 {
-    repeated bytes wallet_addresses = 1;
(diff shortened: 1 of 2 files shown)


proto/mls/database/intents.proto
```diff
@@ -17,12 +18,29 @@ message SendMessageData {
...
+  repeated string account_addresses = 1;
+}
+    
+// Wrapper around a list of repeated Installation IDs
+message InstallationIds {
+  repeated bytes installation_ids = 1;
+}
+
+// One of an EVM account address or Installation ID
+message AddressOrInstallationId {
```
richardhuaaa, line 32: Should this be pluralized too?

richardhuaaa (approved):

PR #119, opened 2023-12-14
BREAKING CHANGE: Move swift GRPC into its own package

nakajima:
Trying to bring in our protos without needing to worry about all the deps of GRPC.


neekolas:
Why do we need Swift GRPC at all? Don't we get all that from Rust now?

neekolas (approved):

nakajima:
@neekolas You're completely right, thanks for the review.

tuddman:
closing per lack of resources for this.  If we need it, please feel free to re-open.

PR #121, opened 2023-12-19
wallet_address -> account_address

richardhuaaa:
Did not update:
- `private_preferences.proto`
- `keystore.proto`
I assume that changes to these would break existing applications


proto/message_api/v3/mls.proto
```diff
@@ -154,7 +154,7 @@ message RevokeInstallationRequest {
 
 // Get all updates for an identity since the specified time
 message GetIdentityUpdatesRequest {
-  repeated string wallet_addresses = 1;
+  repeated string account_addresses = 1;
```
neekolas, line 157: This one is going to require a corresponding change in `xmtp-node-go`

neekolas (approved):
The ones in the "did not update" column feel like something we can leave. Don't know if it's worth the trouble to manage a breaking change there.

If we did care, the move would be to add a second field and fill both. Then, later, we remove the first and mark the field number as "reserved".

richardhuaaa:
Yep, I agree it's not worth the trouble - if we're ever overhauling these in the future, we can bundle it in then

PR #122, opened 2024-01-06
Add Group Metadata

neekolas:
## Summary

- The proto format for group metadata and associated policies


proto/mls/message_contents/group_metadata.proto
```diff
@@ -0,0 +1,63 @@
...
+  CONVERSATION_TYPE_GROUP = 1;
+  CONVERSATION_TYPE_DM = 2;
+}
+
+// The set of policies that govern the group
+message PolicySet {
+  MembershipPolicy add_member_policy = 1;
+  MembershipPolicy remove_member_policy = 2;
+  MembershipPolicy add_installation_policy = 3;
+  MembershipPolicy remove_installation_policy = 4;
```
richardhuaaa, line 28: Is there any reason the add/remove installation policies would be different between different conversations? Shouldn't this be considered baked into the protocol itself, rather than configurable?

proto/mls/message_contents/group_metadata.proto
```diff
@@ -0,0 +1,63 @@
...
+}
+
+// A policy that governs adding/removing members or installations
+message MembershipPolicy {
+  // Base policy
+  enum BasePolicy {
+    BASE_POLICY_UNSPECIFIED = 0;
+    BASE_POLICY_ALLOW = 1;
+    BASE_POLICY_DENY = 2;
+    BASE_POLICY_ALLOW_SAME_MEMBER = 3;
```
richardhuaaa, line 38: This would only be used for removing members right? Is there any case where it would be permissible for this to NOT be set for removing members?

proto/mls/message_contents/group_metadata.proto
neekolas: So, my thinking here is that we should give ourselves flexibility on the read side but not on the write side.

We'll have some prebaked combination of these policies that are the only options available to users (likely one to start). But if we want to roll out new prebaked options (for example, a chat where anyone can add members), we don't have to make a breaking change to the proto and wait for all clients to upgrade in order to roll it out. Only the group creator's app would need to be upgraded to the version with that new option available.

proto/mls/message_contents/group_metadata.proto
neekolas: Probably not. Think that one would be pretty uniform.

proto/mls/message_contents/group_metadata.proto
richardhuaaa: Having the flexibility on the read side means that the write side can be modified by anyone (regardless of if we expose it via our SDK, they can set the payload themselves). Because installation adding/removing is tied in with our security and reliability guarantees, it feels unpredictable to me what can be done by tampering with these fields. Do you have any examples of how we might want to do adding/removing different for installations specifically in the future? I think the example you have is for adding/removing members which totally makes sense.

proto/mls/message_contents/group_metadata.proto
richardhuaaa: I think having this specified as a policy, rather than baked directly into the code, means a malicious person could potentially create a group with this option turned off (no self removal allowed)? I guess my instinct with these changes is to expose the minimum amount for developers to have the features they desire, rather than having maximum flexibility

neekolas:
@richardhuaaa I've removed the `not_condition`, since it was unused and has no immediate use-case.

proto/mls/message_contents/group_metadata.proto
neekolas: OK. I see where you're coming from. Given that we don't _need_ configurability for installation add/remove, why allow for more possible diversity that would be hard to debug (even if attacks using this approach would be very unlikely).

I could be on board with making policies user-configurable only for add/remove members. We'd want to be super-sure there aren't any possibly valid cases where we'd want to change add/remove installations behaviour, but I can't think of any right now.

neekolas:
@richardhuaaa I've removed the Add/Remove installation policies, so the policies are just for members right now.

richardhuaaa (approved):

PR #123, opened 2024-01-08
fix: assume all KeyPackages are last resort

Bren2010:


Bren2010:
The test failure doesn't seem to be related to the PR(?)

richardhuaaa (approved):

neekolas:
Something is off with the Buf workflow config. Not sure how it worked previously, but the action now expects the token in a different field. Will change in a separate PR.

PR #125, opened 2024-01-11
feat: add legacy v2-signed keys

richardhuaaa:
Adds the option for v3 keys to be signed by v2 keys, to avoid additional wallet signatures on clients that already have a signed v2 key.


proto/mls/message_contents/association.proto
```diff
@@ -21,6 +23,14 @@ message Eip191Association {
   string iso8601_time = 4;
 }
 
+// LegacyCreateIdentityAssociation is used when a v3 installation key
+// is signed by a v2 identity key, which in turn is signed via a
+// 'CreateIdentity' wallet signature
+message LegacyCreateIdentityAssociation {
+  RecoverableEcdsaSignature signature = 1;
+  SignedPublicKey legacy_create_identity_key = 2;
```
richardhuaaa, line 31: The creation time and account address can be derived from the nested `SignedPublicKey` proto

neekolas (approved):
Looks good to me

richardhuaaa:
@neekolas, I added https://github.com/xmtp/proto/pull/125/commits/059ce564b59226853a63ca523c0da213db74f442, because I think the revocation must apply to the unsigned legacy identity key bytes, not a proto serialization of the signed key bytes. This is because the proto serialization could be different between implementations. Will land, but let me know if you have feedback or tweaks and I can put up another PR!

neekolas:
> because I think the revocation must apply to the unsigned legacy identity key bytes, not a proto serialization of the signed key bytes. This is because the proto serialization could be different between implementations. Will land, but let me know if you have feedback or tweaks and I can put up another PR!

That's good thinking. We don't want to rely on a byte-perfect proto serialization anywhere.

PR #127, opened 2024-01-11
Add support for query and subscribe of messages via MLS service

snormore:
- Add MLS service query methods and types; QueryGroupMessages and QueryWelcomeMessages
- Replace MLS _tls_serialized fields in messages types with real fields
- Add MLS service subscribe/follow methods and types; FollowGroupMessages and FollowWelcomeMessages
- Rename PublishToGroup to SendGroupMessages
- Rename PublishWelcomes to SendWelcomeMessages


proto/mls/api/v1/mls.proto
```diff
@@ -78,15 +78,47 @@ service MlsApi {
...
+  // Query stored group messages
+  rpc QueryWelcomeMessages(QueryWelcomeMessagesRequest) returns (QueryWelcomeMessagesResponse) {
+    option (google.api.http) = {
+      post: "/mls/v1/query-welcome-messages"
+      body: "*"
+    };
+  }
+
+  // Follow stream of new group messages
+  rpc FollowGroupMessages(FollowGroupMessagesRequest) returns (stream xmtp.mls.message_contents.GroupMessage) {
```
neekolas, line 99: Why follow instead of subscribe? Are any semantics different, or do we just think it's a better name?

snormore: Yea same semantics just different name

snormore: Changed this back to Subscribe per our conversation IRL

proto/mls/message_contents/message.proto
```diff
@@ -22,7 +24,10 @@ message WelcomeMessage {
 message GroupMessage {
   // Version 1 of the GroupMessage format
   message V1 {
-    bytes mls_message_tls_serialized = 1;
+    uint64 id = 1;
```
neekolas, line 27: I noticed this proto is being used in the `SendGroupMessages` request. The caller won't know either the `id` or the `created_ns` value at the time of sending. Is it expected that they set those to 0 values? Or should we have a separate message type for new/unsent messages?

snormore: Will add another type for the query input

proto/mls/api/v1/mls.proto
```diff
@@ -78,15 +78,47 @@ service MlsApi {
...
+    option (google.api.http) = {
+      post: "/mls/v1/query-welcome-messages"
+      body: "*"
+    };
+  }
+
+  // Subscribe stream of new group messages
+  rpc SubscribeGroupMessages(SubscribeGroupMessagesRequest) returns (stream xmtp.mls.message_contents.GroupMessage) {
+    option (google.api.http) = {
+      post: "/mls/v1/follow-group-messages"
```
richardhuaaa, line 101: Small nit, any reason the endpoints are named differently from the protos? E.g. 'follow-group-messages' vs 'SubscribeGroupMessages' here and below, and 'publish' vs 'send' above

proto/mls/api/v1/mls.proto
```diff
@@ -190,3 +222,51 @@ message GetIdentityUpdatesResponse {
...
+  SORT_DIRECTION_UNSPECIFIED = 0;
+  SORT_DIRECTION_ASCENDING = 1;
+  SORT_DIRECTION_DESCENDING = 2;
+}
+
+// Pagination config for queries
+message PagingInfo {
+  SortDirection direction = 1;
+  uint32 limit = 2;
+  uint64 cursor = 3;
```
richardhuaaa, line 237: Does the cursor point to `id` or `created_ns`? Would it make sense to embed that into the name, e.g. `id_cursor`?

proto/mls/message_contents/message.proto
```diff
@@ -10,7 +10,9 @@ option java_package = "org.xmtp.proto.mls.message.contents";
 message WelcomeMessage {
   // Version 1 of the WelcomeMessage format
   message V1 {
-    bytes welcome_message_tls_serialized = 1;
+    uint64 id = 1;
+    uint64 created_ns = 2;
```
richardhuaaa, line 14: Not important, but any reason we need both `id` and `created_ns`? The `created_ns` is server-generated and can't be trusted right?

richardhuaaa (approved):

proto/mls/message_contents/message.proto
neekolas: We do need some timestamp to show to the user, even if we don't trust it for protocol operation.

proto/mls/api/v1/mls.proto
```diff
@@ -190,3 +222,51 @@ message GetIdentityUpdatesResponse {
...
+
+// Response for welcome message queries
+message QueryWelcomeMessagesResponse {
+  repeated xmtp.mls.message_contents.WelcomeMessage messages = 1;
+  PagingInfo paging_info = 2;
+}
+
+// Request for following group messages
+message SubscribeGroupMessagesRequest {
+  bytes group_id = 1;
```
neekolas, line 266: This needs to allow for subscribing to multiple groups so that we can implement some equivalent of `StreamAllMessages`

proto/mls/api/v1/mls.proto
```diff
@@ -78,15 +78,47 @@ service MlsApi {
...
+  // Query stored group messages
+  rpc QueryWelcomeMessages(QueryWelcomeMessagesRequest) returns (QueryWelcomeMessagesResponse) {
+    option (google.api.http) = {
+      post: "/mls/v1/query-welcome-messages"
+      body: "*"
+    };
+  }
+
+  // Subscribe stream of new group messages
+  rpc SubscribeGroupMessages(SubscribeGroupMessagesRequest) returns (stream xmtp.mls.message_contents.GroupMessage) {
```
neekolas, line 99: We've previously been using `Subscribe2` for subscriptions. That requires bidirectional streaming to update the request without ending/restarting the subscription.

proto/mls/api/v1/mls.proto
snormore: Good catch!

proto/mls/api/v1/mls.proto
snormore: Using the `id` - will rename to `id_cursor` :+1:

proto/mls/api/v1/mls.proto
snormore: What do you think about supporting subscribe from cursor instead of bidirectional streaming? IIRC that's why we added it anyway right, to mitigate the issue of missing messages during end/restart?

I'm not sure we'll have bidirectional streaming with the gateway's JSON RPC, and subscribe/continue from cursor seems like a more robust way of dealing with that issue anyway because it works for HTTP too.

neekolas: Hey, if you can implement subscribe-from-cursor across multiple topics in a reasonably performant way...I'm all for it.

proto/mls/api/v1/mls.proto
```diff
@@ -78,15 +78,47 @@ service MlsApi {
...
+    };
+  }
 }
 
-// Publish a batch of MLS messages
-message PublishToGroupRequest {
-  repeated xmtp.mls.message_contents.GroupMessage messages = 1;
+// Send a batch of MLS messages
+message SendGroupMessagesRequest {
+  repeated bytes messages = 1;
```
neekolas, line 117: So, if we make some sort of breaking change to the payload format here, we are comfortable moving to a `SendGroupMessagesV2Request` and a new API? Without the wrapper proto that seems necessary.

I don't have a strong preference. We either end up having to version the payload or the endpoint.

snormore: Hmm ok I'll define a versioned type for it so it can be consistent with the full GroupMessage versioning

proto/mls/api/v1/mls.proto
richardhuaaa: Love this! Would significantly improve reliability and help simplify client logic too

neekolas (approved):

PR #129, opened 2024-01-11
Add `sender_mac` to MessageV2

rygine:


proto/message_contents/message.proto
```diff
@@ -45,6 +45,8 @@ message MessageV2 {
   bytes header_bytes = 1; // encapsulates encoded MessageHeaderV2
   // Ciphertext.payload MUST contain encrypted SignedContent
   Ciphertext ciphertext = 2;
+  // HMAC key derived from the sender's private key
```
rygine, line 48: is this a proper description?

neekolas: I would say `MAC of the message ciphertext, with the MAC key derived from the topic key`

richardhuaaa (approved):

PR #131, opened 2024-01-16
Add legacy v2-signed keys

richardhuaaa:
- Add LegacyCreateIdentityAssociation
- Split out GrantMessagingAccessAssociation and RevokeMessagingAccessAssociation from Eip191Association - not all EIP191 associations will have the same data, and this allows for stronger type-checking
- Change time format from ISO8601 string to nanoseconds since epoch - this creates consistency with the old SignedPublicKey type. The ISO8601 string can be constructed at the time of generating the signature text.

Will land https://github.com/xmtp/libxmtp/pull/425 immediately after this PR lands


neekolas (approved):
This LGTM

proto/mls/message_contents/credential.proto
```diff
@@ -12,14 +12,18 @@ option java_package = "org.xmtp.proto.mls.message.contents";
...
+    LegacyCreateIdentityAssociation legacy_create_identity = 3;
   }
 }
 
 // A declaration and proof that a credential is no longer valid
 message CredentialRevocation {
-  bytes installation_public_key = 1;
+  oneof public_key {
+    bytes installation_key = 1;
+    bytes unsigned_legacy_create_identity_key = 2;
```
neekolas, line 24: What do we mean by unsigned here?

richardhuaaa: This is referring to the `key_bytes` field on `SignedPublicKey`, rather than a serialization of the whole `SignedPublicKey`. We need to do this because proto serialization is not always consistent, so comparing the bytes of the proto may not always match.

Can see why this might be confusing, open to suggestions!

neekolas: Do we care about serialization consistency here? If we've included the exact bytes that were signed, can't we just verify that it deserializes into the right proto. Or are we re-serializing it somewhere?

It seems like it would be helpful to have the wallet signature on the proto, since then you can verify that the wallet doing the revoking matches the wallet that signed the key.

neekolas: Totally possible I'm misunderstanding something

richardhuaaa: > If we've included the exact bytes that were signed, can't we just verify that it deserializes into the right proto

There's two reasons why we might want to attach the unsigned bytes rather than the signed bytes:

1) The client can just do `==` on the unsigned bytes, whereas the alternative is that the client must deserialize the signed bytes, and then do `==` on the unsigned bytes anyway. With the latter, the added complexity means clients might screw this up (e.g., do `==` on the signed bytes)
2) Attaching the unsigned bytes results in a smaller payload that only contains what is needed

> It seems like it would be helpful to have the wallet signature on the proto, since then you can verify that the wallet doing the revoking matches the wallet that signed the key.

I don't think that's the important thing to verify here. Note that a malicious user could sign an existing installation key from someone else using their own wallet, and then revoke it using their own wallet, and then the revocation payload will be internally consistent. I think the important thing to verify here is that within the context of a conversation, the account address on the credential in the conversation matches the account address on the revocation signature.

We're simply saying that this key cannot perform actions on behalf of this account, regardless of how the key was originally registered.

richardhuaaa: I do think maybe there's an issue with this field being ambiguous or confusing. I can add an explanatory comment for what it's referring to, or use a different name if you have any thoughts there?

neekolas: OK. That sounds reasonable. You've convinced me.

PR #132, opened 2024-01-18
Add `GetConversationHmacKeysResponse`

rygine:
this proto is for retrieving HMAC keys for conversation topics from the keystore

### DIFF
--- a/proto/keystore_api/v1/keystore.proto
+++ b/proto/keystore_api/v1/keystore.proto
@@ -295,11 +295,14 @@ message TopicMap {
   map<string, TopicData> topics = 1;
 }
 
+// A mapping of topics to their HMAC keys
 message GetConversationHmacKeysResponse {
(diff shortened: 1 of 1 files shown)


proto/keystore_api/v1/keystore.proto
```diff
@@ -294,3 +294,17 @@ message TopicMap {
   }
   map<string, TopicData> topics = 1;
 }
+
+// A mapping of topics to their HMAC keys
+message GetConversationHmacKeysResponse {
+  // HmacKeyData wraps the HMAC key and the number of 30 day periods since epoch
+  message HmacKeyData {
+    int64 thirty_day_periods_since_epoch = 1;
```
neekolas, line 302: This could probably be an `int32` or even an `int16`, since we know it's going to be a low number. We're still in the hundreds after 50 years

neekolas (approved):

PR #137, opened 2024-01-23
Update post-commit action

neekolas:
## Summary

- With the new `hpke_public_key` we need to store both the `installation_id` and the `hpke_public_key` for each recipient of a Welcome. This updates the data model to support that. It's a breaking change to the proto, but that's fine since there are no long-lived DBs right now.

### DIFF
--- a/proto/mls/database/intents.proto
+++ b/proto/mls/database/intents.proto
@@ -18,11 +18,12 @@ message SendMessageData {
     V1 v1 = 1;
   }
 }
+
 // Wrapper around a list af repeated EVM Account Addresses
 message AccountAddresses {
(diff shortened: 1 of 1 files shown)


proto/mls/database/intents.proto
```diff
@@ -62,9 +63,14 @@ message RemoveMembersData {
 
 // Generic data-type for all post-commit actions
 message PostCommitAction {
+  // An installation
+  message Installation {
+    bytes installation_id = 1;
```
snormore, line 68: Want to rename this to `installation_key` to be consistent with the rename we did in https://github.com/xmtp/proto/pull/127?

snormore (approved):

PR #145, opened 2024-02-12
Bring Beta back into Main

nplasterer:
Android needs the latest proto code to be able to build iOS entitlements.


nplasterer:
@neekolas I can remove the frames stuff from this if you don't want it being released on proto main. The way android works is it would find that new proto code anyways. So it wont be included.

rygine (approved):

neekolas (approved):

PR #146, opened 2024-02-13
feat: Adds AllowGroup and DenyGroup messages

tuddman:
## Summary

- renames `Allow` to `AllowDM`
- renames `Block` to `DenyDM`
- adds `AllowGroup` message type
- adds `DenyGroup` message type


proto/message_contents/private_preferences.proto
```diff
@@ -15,13 +15,19 @@ option java_package = "org.xmtp.proto.message.contents";
...
+    // Add the given group_ids to the allow list
+    repeated string group_ids = 2;
   }
-  // Add the given wallet addresses to the block list
+  // Deny access
   message Block {
+    // Add the given wallet addresses to the block list
     repeated string wallet_addresses = 1;
+    // Add the given group_ids to the block list
+    repeated string group_ids = 2;
```
neekolas, line 30: I think we are going to want a separate action for these, rather than a second field in the existing action. Otherwise you could specify both in the same action, which is confusing. 

We can safely rename the type names (although all our SDKs will have to update their import statements)

tuddman: Sure.  I updated to pull out `AllowGroup` and `BlockGroup` into their own message types.
and renamed `Allow` to `AllowDM`  and `Block` to `BlockDM`.  I'll update the SDK side here next.

proto/message_contents/private_preferences.proto
```diff
@@ -11,28 +11,42 @@ import "message_contents/ciphertext.proto";
...
+  // Deny 1:1 direct message (DM) access
+  message BlockDM {
+    // Add the given wallet addresses to the block list
     repeated string wallet_addresses = 1;
   }
 
+  // Grant Group access
+  message AllowGroup {
+    // Add the given group_ids to the allow list
+    repeated string group_ids = 2;
```
neekolas, line 32: `group_ids` should `repeated bytes`

neekolas (changes requested):
We've opted for `group_id` to be treated as bytes everywhere

neekolas (approved):

PR #150, opened 2024-02-21
Add MlsEnvelope

neekolas:
## Summary

Adds a new `MlsEnvelope` that wraps both the `EncodedContent` and the `idempotency_key` fields.

It's important to clearly differentiate where this envelope is used, and where we want to break things apart.

Data going over the wire to the API: A serialized `MlsEnvelope` is the result of decrypting the `data` field of a V1 GroupMessage
Data stored in the DB: `content` and `idempotency_key` stored separately
Messages returned to callers of `list` or `stream` methods: `content` is only returned

Generally I don't think the idempotency key should ever make its way over the FFI boundary to integrators. WDYT about that?

## Other names considered
- `Envelope`. We already have a V1 `Envelope`, which has unencrypted fields. Seems confusing
- `EncryptedEnvelope`. Feels unclear whether the contents are encrypted or whether this message is meant to be encrypted
- `InnerEnvelope`. I don't hate this, although it's probably more accurate to say that it's in the middle. `EncodedContent` lives inside of it.

## Issue
https://github.com/xmtp/libxmtp/issues/529


proto/mls/message_contents/content.proto
```diff
@@ -39,4 +39,20 @@ message EncodedContent {
...
-}
\ No newline at end of file
+}
+
+// An MlsEnvelope is the outermost payload that gets encrypted by MLS
+message MlsEnvelope {
+  // Version 1 of the encrypted envelope
+  message V1 {
+    // Expected to be EncodedContent
+    bytes content = 1;
```
richardhuaaa, line 49: Any reason we have to serialize the EncodedContent to bytes here, rather than just nesting it as-is?

richardhuaaa (approved):
> Data stored in the DB: content and idempotency_key stored separately

Do you mean the libxmtp DB here? Could we just use the key to derive and store the message_id, and then throw the key away?

> Generally I don't think the idempotency key should ever make its way over the FFI boundary to integrators. WDYT about that?

That makes a lot of sense!

richardhuaaa:
There's 7k lines removed here - do you need to add the code-genned stuff back?

tuddman:
Screenshot 2024-02-21 at 19 45 44

Normally I love a PR that reduces LoC + increases functionality, but: Is this intentional & correct?

richardhuaaa:
> Other names considered

Maybe `DecryptedMessageEnvelope`, or `DecryptedEnvelope` or `PlaintextEnvelope`? `MlsEnvelope` is fine too

neekolas:
Something must have gone haywire in the go generator. Re-ran it and it added the code back

proto/mls/message_contents/content.proto
neekolas: I was going to post that as a question to the group.

Right now, platform SDKs handle all decoding of content. They expect messages as bytes and decode to `EncodedContent` and then decode the inner content using content types.

While we could nest `EncodedContent` here, we would need to re-serialize the `EncodedContent` when we store it in the DB. And to send a message we'd need to deserialize the `EncodedContent` passed to `libxmtp` by the caller to create this payload.

neekolas:
@richardhuaaa I don't mind `PlaintextEnvelope`

proto/mls/message_contents/content.proto
richardhuaaa: I see! Good catch - fine with leaving it as bytes.

tuddman:
> @richardhuaaa I don't mind `PlaintextEnvelope`

👍 That's the clearest by my reading in describing what this is.

tuddman (approved):

PR #152, opened 2024-03-21
Add message backup request and response

tuddman:
## Summary

Initial stab at adding support for Message Backups.


proto/mls/message_contents/content.proto
```diff
@@ -52,7 +52,35 @@ message PlaintextEnvelope {
...
+  message V2 {
+    oneof message_type {
+      // Expected to be EncodedContent
+      bytes content = 1;
+      // Initiator sends the request to receive history
+      MessageHistoryRequest message_history_request = 2;
+      // Other credentialed party sends response
+      MessageHistoryResponse message_history_response = 3;
+    }
+  }
```
nplasterer, line 65: Should we be adding message backups for V1 messages as well?

nplasterer (approved):

proto/mls/message_contents/content.proto
tuddman: Perfectly good question!  that I don't know the answer to.  
This is very much a first pass to get a basic round-trip request<>response functioning. 
If we need to backport and/or add additional support for `V1` as well, we can absolutely do that.

PR #155, opened 2024-03-28
Prototype identity apis

neekolas:
## Summary

I wanted to sketch out what the APIs would look like to support the association event log. Implemented as a new GRPC service with new Protobufs for the associations.


proto/identity/associations/association.proto
```diff
@@ -0,0 +1,78 @@
...
+  enum AssociatedRole {
+    ASSOCIATED_ROLE_UNSPECIFIED = 0;
+    ASSOCIATED_ROLE_INSTALLATION = 1;
+    ASSOCIATED_ROLE_ADDRESS = 2;
+    ASSOCIATED_ROLE_LEGACY_KEY = 3;
+  }
+
+  string xid = 1;
+  uint64 client_timestamp_ns = 2;
+  uint32 nonce = 3;
```
richardhuaaa, line 30: If the nonce is meant to be globally unique, a string GUID might be better.

However, I think it's better if it is not required to be globally unique - for replay protection, we really just need it to be unique across the (existing member, new member, role) tuple. If that's the case, could we use the `client_timestamp_ns` as effectively the nonce? It has the benefit that unintentional nonce reuse is very hard to do, as well as the fact that it displays better in association text to the user.

And instead of storing the `revoked_nonce` below, I wonder if it makes sense to store a hash of the whole association instead. That way, it is clear that there is no assumption of global uniqueness for the nonce.

proto/identity/associations/association.proto
```diff
@@ -0,0 +1,78 @@
...
+  Signature recovery_address_signature = 4;
+  // The nonce of the association being removed.
+  // This nonce must never be used again for this member.
+  uint32 revoked_nonce = 4;
+  // Either an installation public key or a wallet address
+  string revoked_member = 5;
+  // Allow the revoker to include a list of downstream associations that should
+  // be retained even after the revocation. Any association not in this list
+  // will be removed
+  repeated string retained_associations = 6;
```
richardhuaaa, line 51: What are these strings? I think a hash of each association makes the most sense here.

proto/identity/associations/association.proto
```diff
@@ -0,0 +1,78 @@
...
+// An entity may have been associated with multiple roles. Clients MUST remove
+// the entity from all previously associated roles.
+message RevokeAssociation {
+  string xid = 1;
+  uint64 client_timestamp_ns = 2;
+  uint32 nonce = 3;
+  Signature recovery_address_signature = 4;
+  // The nonce of the association being removed.
+  // This nonce must never be used again for this member.
+  uint32 revoked_nonce = 4;
```
richardhuaaa, line 45: Could this be a hash of the whole association?

richardhuaaa (commented):
One thing I'm wondering is if we can add the capability to batch sign things.

For example, a 'move XID' operation is a revoke followed by an add. A 'register XID' operation is potentially a create followed by a change recovery address. Can we just concatenate the text and sign the result? Can the API support all identity updates via a single endpoint?

proto/identity/associations/association.proto
richardhuaaa: I guess if we used timestamps instead of nonces/hashes, for this field and the `retained_associations` field, the signature text would be more comprehensible to the user. Not sure how much this matters

```
Revoke 0xabcdef
Which was created at: <timestamp>
The following associations will remain valid:
0x234324 created at: <timestamp>
0x656734 created at: <timestamp>
```

proto/identity/associations/association.proto
neekolas: I was imagining the IDs of the entities (wallet address or installation pub key), but the hash of the association would be fine too.

proto/identity/associations/association.proto
neekolas: I do like the human readability of it being a timestamp.

proto/identity/associations/association.proto
neekolas: The nonce only needs to be locally unique to the XID. Timestamp would work, but I'm imagining it's going to get generated somewhere inside our SDK so I'm not sure it matters.

> And instead of storing the revoked_nonce below, I wonder if it makes sense to store a hash of the whole association instead. That way, it is clear that there is no assumption of global uniqueness for the nonce.

Works for me

proto/identity/associations/association.proto
```diff
@@ -0,0 +1,70 @@
...
+  AssociatedRole new_member_role = 6;
+}
+
+// An entity may have been associated with multiple roles. Clients MUST remove
+// the entity from all previously associated roles.
+message RevokeAssociation {
+  string xid = 1;
+  uint64 client_timestamp_ns = 2;
+  uint32 nonce = 3;
+  Signature recovery_address_signature = 4;
```
neekolas, line 42: Good catch on making this a one-way signature. Total mistake on my part

proto/identity/associations/association.proto
```diff
@@ -0,0 +1,70 @@
...
+
+message AddAssociation {
+  enum AssociatedRole {
+    ASSOCIATED_ROLE_UNSPECIFIED = 0;
+    ASSOCIATED_ROLE_INSTALLATION = 1;
+    ASSOCIATED_ROLE_ADDRESS = 2;
+    ASSOCIATED_ROLE_LEGACY_KEY = 3;
+  }
+
+  string xid = 1;
```
neekolas, line 28: The reason I had this field in the API request instead of here is because I'm imagining these whole entries are going to get serialized into the credential and stored in the DB. Less duplication that way, although it's obviously a small optimization in the scheme of things.

proto/identity/associations/association.proto
```diff
@@ -0,0 +1,70 @@
...
+  uint64 client_timestamp_ns = 2;
```
neekolas, line 29: What is the `client_timestamp_ns` used for? Seems dangerous to trust it for anything other than using it as a nonce.

proto/identity/associations/association.proto
```diff
@@ -0,0 +1,70 @@
...
+
+import "identity/associations/signature.proto";
+
+option go_package = "github.com/xmtp/proto/v3/go/identity/associations";
+option java_package = "org.xmtp.proto.identity.associations";
+
+// The first entry of any XID log, which is considered valid iff the XID can be
+// deterministically derived from the address and nonce.
+// The recovery address defaults to the initial_associated_address unless
+// there is a subsequent ChangeRecoveryAddress in the log.
```
neekolas, line 13: I still can't get my head around how unsigned XID creation that's associated with a wallet will work. We can chat tomorrow. Maybe I'm missing something.

proto/identity/associations/association.proto
richardhuaaa: Removed these as the design no longer needs them

proto/identity/associations/association.proto
richardhuaaa: Honestly, I think both server timestamps and client timestamps are untrustworthy, depending on the trust model. I think it is helpful to assume that any consumer will account for this, and to include both and explicitly label where they came from.

The uses for the client timestamp here could be: as a nonce, as a human readable value in the signature that the user can verify, as a piece of metadata that the user can use to match an installation to an app they have installed.

proto/identity/associations/association.proto
richardhuaaa: Okay, so I removed the nonce. For replay protection, I'd suggest we detect replays using the hash of the whole association - it is up to the client to ensure that the client timestamp will be different between different signature attempts if they want their signature to be accepted.

proto/identity/associations/association.proto
richardhuaaa: Updated the XID creation to allow for unsigned creation with a nonce of 0, as well as signed creation with other nonces

proto/identity/associations/association.proto
```diff
@@ -0,0 +1,68 @@
...
+  }
+
+  string xid = 1;
+  uint64 client_timestamp_ns = 2;
+  AssociatedRole new_member_role = 3;
+  Signature existing_member_signature = 4;
+  Signature new_member_signature = 5;
+}
+
+// An entity may have been associated with multiple roles. Clients MUST remove
```
neekolas, line 43: When and how would an entity be associated with multiple roles?

richardhuaaa: You could potentially associate the same keypair as a wallet address AND an installation key. I suppose we can just prohibit that

neekolas: Makes life easier if we can keep it to one. Then we can prefix all the IDs with their role when creating keys for the hashmap and categorically prevent collisions.

proto/identity/associations/association.proto
richardhuaaa: Updated the XID creation to require no independent signature, but must be batched with an AddAssociation.

proto/identity/associations/association.proto
richardhuaaa: Are you referring to the server DB or client DB? And with the new design, credentials will only contain an XID, sequence number, and public installation key right?

My main thinking with placing the XID here is that for constructing the signature text, all of the information should be in one place

neekolas: I was thinking about the server DB records, since the xid is going to be stored outside of this payload somewhere so it can be indexed. 

My comment about the credential is out of date. I do think all we need in the credential are the things you listed above.

proto/identity/associations/association.proto
```diff
@@ -0,0 +1,59 @@
...
+// the entity from all previously associated roles.
+message RevokeAssociation {
+  string xid = 1;
+  uint64 client_timestamp_ns = 2;
+  // Either an installation or wallet public key
+  bytes revoked_public_key = 3;
```
neekolas, line 40: I don't think we can assume all entities have public keys. For example, Smart Wallets don't.

proto/identity/associations/association.proto
richardhuaaa: @neekolas I just realized that protobuf serialization is not guaranteed to be consistent, hence we can't use a hash of the protobuf bytes for replay protection. Maybe a hash of the signature bytes would work? If not, maybe we could manually construct a tuple with our own serialization method

neekolas: In my implementation I've just baked a function `hash()` into the trait for each log entry, so each type can define an appropriate hash function.

I haven't worked out the signature text for all the log entries, but I _think_ that's the only input needed to get replay prevention. Anything important should go into the signature text.

If that's the case I'll remove the `hash()` function and just use the text.

proto/identity/associations/association.proto
```diff
@@ -0,0 +1,87 @@
...
+    ASSOCIATED_ROLE_UNSPECIFIED = 0;
+    ASSOCIATED_ROLE_INSTALLATION = 1;
+    ASSOCIATED_ROLE_ADDRESS = 2;
+  }
+
+  string xid = 1;
+  uint64 client_timestamp_ns = 2;
+  AssociatedRole new_member_role = 3;
+  Signature existing_member_signature = 4;
+  Signature new_member_signature = 5;
```
neekolas, line 37: There's a circular dependency here with generating the signature text. 

We need the new member's address or public key to generate the text, but we need the text to recover an identity out of most of these signature types.

neekolas: The easy fix is to at least include the identifier for the new entity. Maybe the existing one too

proto/identity/associations/signature.proto
```diff
@@ -0,0 +1,57 @@
...
+  bytes signature = 3;
+}
+
+// An existing address on xmtpv2 may have already signed a legacy identity key
+// of type SignedPublicKey via the 'Create Identity' signature.
+// For migration to xmtpv3, the legacy key is permitted to sign on behalf of the 
+// address to create a matching xmtpv3 installation key.
+// This signature type can ONLY be used for CreateXid and AddAssociation
+// payloads, and can only be used once in xmtpv3.
+message LegacyDelegatedSignature {
```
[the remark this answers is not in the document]
neekolas: We've decided to go with the first mitigation, and will only allow one action to be done with the legacy key per XID. All subsequent actions will be rejected.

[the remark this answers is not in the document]
richardhuaaa: Summary of discussion - no matter what, we need to ensure that the legacy key can only be used once. We can bake this into the replay prevention logic. That way, there's no need to revoke it - during revocation, the user either revokes the installation or the address, without having to worry about the legacy key.

richardhuaaa:
By the way, do we have a preferred method of versioning/future-proofing these associations? Is that even possible?

proto/identity/api/v1/identity.proto
```diff
@@ -51,31 +50,35 @@ message PublishIdentityUpdateRequest {
...
+  string inbox_id = 1;
+}
+
 // Get all updates for an identity since the specified time
 message GetIdentityUpdatesRequest {
   // Points to the last entry the client has received. The sequence_id should be
   // set to 0 if the client has not received anything.
-  message XidCursor {
-    string xid = 1;
+  message InboxCursor {
```
richardhuaaa, line 61: Should we actually just call this `Request`, like on the other endpoints? I fear `InboxCursor` might be interpreted as a cursor to the messages in your inbox.

proto/identity/api/v1/identity.proto
```diff
@@ -51,31 +50,35 @@ message PublishIdentityUpdateRequest {
   xmtp.identity.associations.IdentityUpdate identity_update = 1;
 }
 
+message PublishIdentityUpdateResponse {
+  string inbox_id = 1;
```
richardhuaaa, line 54: Why is this needed? Is the client expected to validate this result?

neekolas: A `CreateIdentity` request doesn't contain the `inbox_id` explicitly, just the inputs. Just gives the clients some confirmation that the server came to the same conclusion they did.

richardhuaaa: If someone wants to move a wallet between inbox ID's, can they publish a batch signature to revoke from one inbox ID and associate with another, and have it apply to both inbox ID's?

richardhuaaa: Summary of discussion: it is possible to support multi-inbox identity updates in order to reduce the number of signatures a user needs to provide. However, this case should be rare and adds complexity, so we choose not to optimize for it.

richardhuaaa (approved):

PR #156, opened 2024-04-02
feat: Add Frames Protos to Kotlin and Swift

alexrisch:
Updated generate scripts to include Frames Protos

### DIFF
--- a/dev/swift/generate
+++ b/dev/swift/generate
@@ -36,6 +36,7 @@ docker run --rm -i -v${PWD}:/code xmtp/protoc-swift \
   message_contents/private_preferences.proto \
   message_contents/signed_payload.proto \
   message_contents/ecies.proto \
+  message_contents/frames.proto \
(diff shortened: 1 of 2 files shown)


dev/swift/generate
```diff
@@ -36,6 +36,7 @@ docker run --rm -i -v${PWD}:/code xmtp/protoc-swift \
   message_contents/private_preferences.proto \
   message_contents/signed_payload.proto \
   message_contents/ecies.proto \
+  message_contents/frames.proto \
```
neekolas, line 39: I believe Swift no longer uses these scripts and instead generates in `xmtp-ios` using Buf.

alexrisch: That's correct, I updated it for parity, but can remove it

neekolas: We should just axe the whole thing, but can do that in another PR

neekolas (approved):

PR #160, opened 2024-04-10
feat: AssociationState Protos for Validation Service

insipx:


neekolas (approved):

proto/mls_validation/v1/service.proto
```diff
@@ -59,3 +64,24 @@ message ValidateGroupMessagesResponse {
...
+
+// Map of members belonging to an inbox_id
+message MemberMap {
+  xmtp.identity.associations.MemberIdentifier key = 1;
+  xmtp.identity.associations.Member value = 2;
+}
+
+// Reponse to GetAssociationStateRequest, containing the final association state
+// for an InboxID
+message GetAssociationStateResponse {
```
richardhuaaa, line 82: The server will need to know which addresses were affected, and whether they were associations or revocations, for the last appended identity update. What would be the best way to expose this?

richardhuaaa (changes requested):

proto/mls_validation/v1/service.proto
richardhuaaa: https://github.com/xmtp/xmtp-node-go/blob/main/pkg/identity/api/v1/service.go#L53-L56

richardhuaaa: Perhaps an optional repeated field in `GetAssociationStateRequest` for `new_updates` or similar? Open to feedback here

richardhuaaa: I think we would also need this on the client for commit construction and validation: https://community.xmtp.org/t/xip-46-multi-wallet-identity/639#validating-the-other-proposals-in-the-commit-31

Perhaps the `GetAssociationStateRequest` can just have an `old_sequence_id` field, with all changes after that being tracked?

richardhuaaa:
Also, if validation fails, it will fail on a specific sequence number, and I assume you'll abort the rest of the process. Could we get the sequence number that failed, and the state + changes up to that point? This will help with handling concurrency on the server

neekolas (approved):

richardhuaaa (approved):

PR #163, opened 2024-04-11
feat: Consent Proof Proto

alexrisch:
Added consent proof into invitation

### DIFF
--- a/proto/message_contents/invitation.proto
+++ b/proto/message_contents/invitation.proto
@@ -40,6 +40,9 @@ message InvitationV1 {
     // Specify the encryption method to process the key material properly.
     Aes256gcmHkdfsha256 aes256_gcm_hkdf_sha256 = 3;
   }
+
+  // The user's consent proof
+  ConsentProofPayload consent_proof = 4;
 }
 
 // Sealed Invitation V1 Header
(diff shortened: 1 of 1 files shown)


nplasterer (approved):

proto/message_contents/invitation.proto
```diff
@@ -70,3 +73,14 @@ message SealedInvitation {
...
+
+// Payload for user's consent proof to be set in the invitation
+// Signifying the conversation should be preapproved for the user on receipt
+message ConsentProofPayload {
+  // the user's signature in hex format
+  string signature = 1;
+  // approximate time when the user signed
+  uint64 timestamp = 2;
+  // version of the payload
+  uint32 payload_version = 3;
```
neekolas, line 85: I don't have a strong preference here, but it might be nicer to have an Enum for the payload version. That's what we do in other places, even if the only option right now is `VERSION_ONE`

neekolas (approved):

proto/message_contents/invitation.proto
```diff
@@ -70,3 +73,18 @@ message SealedInvitation {
     SealedInvitationV1 v1 = 1;
   }
 }
+
+enum ConsentProofPayloadVersion {
+  V1 = 1;
```
neekolas, line 78: We have a proto lint rule that forces you to have a prefix of the enum name on each value and to define an `$PREFIX_UNSPECIFIED` values as the `0`

neekolas (approved):

PR #167, opened 2024-04-16
Add group membership proto

neekolas:
## tl;dr

Add `GroupMembership` proto in line with [XIP-46](https://community.xmtp.org/t/xip-46-multi-wallet-identity/639)

### DIFF
--- a/proto/mls/message_contents/group_membership.proto
+++ b/proto/mls/message_contents/group_membership.proto
@@ -1,0 +1,13 @@
+// Group metadata
+syntax = "proto3";
+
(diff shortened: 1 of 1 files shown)


nplasterer (approved):
Kind of a pain but for this to get picked up by kotlin in the maven package you need to add the new file to this long list https://github.com/xmtp/proto/blob/main/dev/kotlin/generate#L53 Alex looked into allowing it to take a wild card but wasn't successful.

37ng (approved):

PR #172, opened 2024-04-23
Revert "feat: update group mutable metadata to use map for upgradability"

cameronvoell:
Reverts xmtp/proto#171


cameronvoell:
@insipx maybe we revert this until your libxmtp PR lands

PR #173, opened 2024-04-23
update proto for signature verification

37ng:
Will merge this after [signature verification](https://github.com/xmtp/libxmtp/pull/673) in libxmtp is merged.

### DIFF
--- a/proto/identity/associations/signature.proto
+++ b/proto/identity/associations/signature.proto
@@ -16,7 +16,10 @@ message RecoverableEcdsaSignature {
 
 // EdDSA signature for 25519
 message RecoverableEd25519Signature {
+  // 64 bytes [R(32 bytes) || S(32 bytes)]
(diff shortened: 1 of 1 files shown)


proto/identity/associations/signature.proto
```diff
@@ -28,6 +31,8 @@ message Erc1271Signature {
   uint64 block_number = 2;
   // The actual signature bytes
   bytes signature = 3;
+  // RPC URL of the node where the contract is deployed
+  string chain_rpc_url = 4;
```
neekolas, line 35: Usually with RPC providers this would be handled at the app level, since the RPC URL might include secrets that are scoped to the app (like an Infura API key).

These signatures are public as part of your Inbox Log, and I don't think apps are going to be happy sharing these kinds of RPC URLs amongst each other.

neekolas (approved):

PR #177, opened 2024-04-26
Inbox ID credential proto

37ng:


proto/mls/message_contents/credential.proto
```diff
@@ -8,15 +8,6 @@ import "mls/message_contents/association.proto";
 option go_package = "github.com/xmtp/proto/v3/go/mls/message_contents";
 option java_package = "org.xmtp.proto.mls.message.contents";
 
-// A credential that can be used in MLS leaf nodes
```
neekolas, line 11: If we're putting this into main, let's just leave this old credential as-is and add the new one. That way it's not a breaking change

37ng: I rebased this to `identity` branch, deleted part is for CI to pass.

proto/identity/credential.proto
```diff
@@ -0,0 +1,12 @@
+// Credentials
+syntax = "proto3";
+
+package xmtp.identity;
+
+option go_package = "github.com/xmtp/proto/v3/go/mls/message_contents";
```
neekolas, line 6: The `go_package` should be `github.com/xmtp/proto/v3/go/identity`

proto/identity/credential.proto
```diff
@@ -0,0 +1,12 @@
...
+option java_package = "org.xmtp.proto.mls.message.contents";
```
neekolas, line 7: This should be `org.xmtp.proto.identity`

proto/mls/message_contents/credential.proto
neekolas: I still think we might want to keep the old one around for a bit. Once we start using this in `libxmtp` we are going to have to update every single reference to the old credential immediately or compiles will fail.

neekolas (approved):

PR #178, opened 2024-04-29
Protos for credential validation of Inbox IDs

insipx:
- Credential validation of inbox IDs
- Validate a InboxID/Installation Public Key from a device
- Rel to https://github.com/xmtp/libxmtp/issues/675

### DIFF
--- a/proto/mls_validation/v1/service.proto
+++ b/proto/mls_validation/v1/service.proto
@@ -85,7 +85,7 @@ message GetAssociationStateResponse {
 }
 
 // Request to validate an InboxID with the backend service. Ensures an Inbox Id <> Installation key are valid.
-message ValidateInboxIdRequest {
+message ValidateInboxIdsRequest {
   // a single validation request
   message ValidationRequest {
     xmtp.identity.MlsCredential credential = 1;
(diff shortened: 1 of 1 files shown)


proto/mls_validation/v1/service.proto
```diff
@@ -78,3 +83,28 @@ message GetAssociationStateResponse {
   xmtp.identity.associations.AssociationState association_state = 1;
   xmtp.identity.associations.AssociationStateDiff state_diff = 2;
 }
+
+// Request to validate an InboxID with the backend service. Ensures an Inbox Id <> Installation key are valid.
+message ValidateInboxIdsRequest {
+  // a single validation request
+  message ValidationRequest {
+    xmtp.identity.MlsCredential credential = 1;
```
neekolas, line 91: How does the Rust backend have access to the inbox history needed to validate the credential? That all lives in go-land?

neekolas: I think this method needs to take in all the IdentityUpdates for the inbox as well, then it can be stateless on the Rust side

neekolas (approved):

37ng (approved):

proto/mls_validation/v1/service.proto
```diff
@@ -16,6 +17,26 @@ service ValidationApi {
   
   // Gets the final association state for a batch of identity updates
   rpc GetAssociationState(GetAssociationStateRequest) returns (GetAssociationStateResponse) {}
+  
+  // Validates InboxID key packages and returns credential information for them, without validating anything.
```
37ng, line 21: what does "without validating anything" mean?

PR #180, opened 2024-05-01
Inbox ID updates

neekolas:
## tl;dr

- Adds new `UpdateGroupMembershipData` intent, which will be used for adding/removing members
- Adds `creator_inbox_id` to `GroupMetadataV1` (but doesn't yet remove `creator_account_address`)

### DIFF
--- a/proto/mls/database/intents.proto
+++ b/proto/mls/database/intents.proto
@@ -61,6 +61,22 @@ message RemoveMembersData {
   }
 }
 
+// The data required to make a commit that updates group membership
+// Handles both Add and Remove actions
+message UpdateGroupMembershipData {
+  // V1 of UpdateGroupMembershipPublishData
+  message V1 {
+    // Contains delta of membership updates that need to be applied
+    map<string, uint64> membership_updates = 1;
+    // Contains the list of members that will be removed
+    repeated string removed_members = 2;
+  }
+
+  oneof version {
+    V1 v1 = 1;
+  }
+}
+
 // The data required to update group metadata
 message UpdateMetadataData {
   // V1 of UpdateMetadataPublishData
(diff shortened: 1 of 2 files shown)


proto/mls/message_contents/group_metadata.proto
```diff
@@ -9,8 +9,10 @@ option java_package = "org.xmtp.proto.mls.message.contents";
 // Parent message for group metadata
 message GroupMetadataV1 {
   ConversationType conversation_type = 1;
+  // This will be removed soon
```
tuddman, line 12: Absolute confidence it will be 😛

tuddman (approved):

proto/mls/database/intents.proto
```diff
@@ -61,6 +61,22 @@ message RemoveMembersData {
   }
 }
 
+// The data required to make a commit that updates group membership
+// Handles both Add and Remove actions
+message UpdateGroupMembershipData {
+  // V1 of UpdateGroupMembershipPublishData
+  message V1 {
+    // Contains delta of membership updates that need to be applied
+    map<string, uint64> membership_updates = 1;
```
37ng, line 70: What are the key, value of this map? (Might be good to document it)

proto/mls/database/intents.proto
```diff
@@ -61,6 +61,22 @@ message RemoveMembersData {
...
+    // Contains the list of members that will be removed
+    repeated string removed_members = 2;
```
37ng, line 72: If the `membership_updates` contains delta, what are `removed_member` for?

neekolas: I thought about having an update that has a `sequence_id` of 0 be treated as a remove, but it felt a little hacky.

PR #186, opened 2024-05-18
fix: change inbox id to bytes

nplasterer:
🤦‍♀️ This should have been bytes instead of string.


37ng:
the [inbox_id in libxmtp](https://github.com/xmtp/libxmtp/blob/main/xmtp_id/src/lib.rs#L31) is string, are they supposed to match?

nplasterer:
Ahh okay oops that was my bad I thought it would be a bytearray for some reason 😄

PR #188, opened 2024-05-20
Flags for Key Package Validation protos

insipx:


proto/mls_validation/v1/service.proto
```diff
@@ -35,6 +35,7 @@ message ValidateInboxIdKeyPackagesResponse {
     string error_message = 2;
     xmtp.identity.MlsCredential credential = 3;
     bytes installation_public_key = 4;
+    uint64 expiration = 6;
```
37ng, line 38: should this be 5?

37ng (approved):

PR #189, opened 2024-05-24
Add transactionId to FrameActionBody

daria-github:
Adding for full tx spec support.

### DIFF
--- a/proto/message_contents/frames.proto
+++ b/proto/message_contents/frames.proto
@@ -30,6 +30,8 @@ message FrameActionBody {
   string state = 7;
   // A 0x wallet address
   string address = 8;
+  // A hash from a transaction
+  string transactionId = 9;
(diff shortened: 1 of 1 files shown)


rygine (approved):

proto/message_contents/frames.proto
```diff
@@ -30,6 +30,8 @@ message FrameActionBody {
   string state = 7;
   // A 0x wallet address
   string address = 8;
+  // A hash from a transaction
+  string transaction_id = 9;
```
37ng, line 34: if this is the hash of a eth transaction, I think `transaction_hash` would be a more consistent naming.

daria-github: thanks @37ng! the naming is part of the spec, so need to keep this one as-is.

PR #191, opened 2024-06-12
Update `SmartContractWalletSignature`

37ng:
- Renaming
- Add a `chain_rpc_url` field to it

### DIFF
--- a/proto/identity/associations/signature.proto
+++ b/proto/identity/associations/signature.proto
@@ -22,15 +22,17 @@ message RecoverableEd25519Signature {
   bytes public_key = 2;
 }
 
-// Smart wallet signature
-message Erc1271Signature {
-  // CAIP-10
+// Smart Contract Wallet signature
+message SmartContractWalletSignature {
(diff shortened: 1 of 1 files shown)


proto/identity/associations/signature.proto
```diff
@@ -53,7 +55,7 @@ message Signature {
   //    must be inferred from the context in which the signature is provided.
   oneof signature {
     RecoverableEcdsaSignature erc_191 = 1;
-    Erc1271Signature erc_1271 = 2;
+    SmartContractWalletSignature erc_6492 = 2;
```
richardhuaaa, line 58: The name here is `erc_6492`, but it's not necessarily a 6492 signature right?

richardhuaaa (approved):

proto/identity/associations/signature.proto
37ng: That's right, basically 6492 handles 1271 & EOA signatures as well.

PR #195, opened 2024-07-24
Initial protos for decentralization

richardhuaaa:
Refer to XIP draft for rationale

### DIFF
--- a/proto/xmtpv4/message_api/message_api.proto
+++ b/proto/xmtpv4/message_api/message_api.proto
@@ -1,3 +1,4 @@
+// Message API for XMTP V4
 syntax = "proto3";
 
 package xmtp.xmtpv4;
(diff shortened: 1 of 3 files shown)


neekolas (approved):

richardhuaaa:
Heads up @neekolas, I made a few small tweaks since you last signed off

proto/xmtpv4/message_api/message_api.proto
```diff
@@ -0,0 +1,129 @@
...
+message OriginatorEnvelope {
+  bytes unsigned_originator_envelope = 1; // Protobuf serialized
+  oneof proof {
+    xmtp.identity.associations.RecoverableEcdsaSignature originator_signature = 2;
+    BlockchainProof blockchain_proof = 3;
+  }
+}
+
+// Wraps originator envelope with a sequence ID assigned by the gateway
+message GatewayEnvelope {
```
neekolas, line 54: I wonder if "gateway" is the right term here. When I hear "gateway" my mind immediately associates it with being the first thing that touches the message. In this case it's really the last.

What about `ReplicaEnvelope`?

PR #196, opened 2024-08-01
Add publish envelope rpc, generalize to multiple payload types

richardhuaaa:
- Add publish envelope RPC
- Generalize beyond application message to welcomes, identity updates, etc
- Lock the target originator and topic via the payer signature

### DIFF
--- a/.protolint.yaml
+++ b/.protolint.yaml
@@ -44,6 +44,10 @@ lint:
       - ENUM_FIELDS_HAVE_COMMENT
       # This one doesn't like single word RPCs, which I have no problem with
       - RPC_NAMES_CASE
+      # In practice this creates a lot of meaningless comments, we can trust
+      # authors to determine when comments will be useful or not
+      - MESSAGES_HAVE_COMMENT
+      - SERVICES_HAVE_COMMENT
 
   # Linter rules option.
   rules_option:
--- a/proto/xmtpv4/message_api/message_api.proto
+++ b/proto/xmtpv4/message_api/message_api.proto
@@ -128,19 +128,7 @@ message PublishEnvelopeRequest {
 }
 
 message PublishEnvelopeResponse {
-  message PublishEnvelopeSuccess {
-    OriginatorEnvelope originator_envelope = 1;
-  }
-  message PublishEnvelopeFailure {
-    // TODO. Possible reasons:
-    // last_originator_sids are not synced
-    // identity validation failed, if we are doing this server-side
-  }
-
-  oneof result {
-    PublishEnvelopeSuccess success = 1;
-    PublishEnvelopeFailure failure = 2;
-  }
+  OriginatorEnvelope originator_envelope = 1;
 }
 
 // Replication API
--- a/proto/xmtpv4/message_api/message_api.proto
+++ b/proto/xmtpv4/message_api/message_api.proto
@@ -5,20 +5,32 @@ package xmtp.xmtpv4;
 
 import "google/api/annotations.proto";
 import "identity/associations/signature.proto";
+import "mls/api/v1/mls.proto";
 
 option go_package = "github.com/xmtp/proto/v3/go/xmtpv4/message_api";
 
-// Authenticated data within the MlsPrivateMessage
+// Data visible to the server that has been signed by the client.
+// Included in one of:
+// - MLS PrivateMessage.authenticated_data for messages and commits
+// - MLS KeyPackageTBS.extensions for key packages
+// - Omitted for Welcome messages
 message AuthenticatedData {
   repeated uint64 last_originator_sids = 1;
 }
 
-// Replaces GroupMessageInput V1
-// To rename or not to rename?
+// Data visible to the server that has been signed by the payer. Fields can
+// be set by the client OR the payer, but the final result will be returned
+// to the client after publishing.
 message ClientEnvelope {
-  // TLS serialized MlsMessageIn, which contains MlsPrivateMessage
-  bytes data = 1;
-  bytes sender_hmac = 2;
+  uint32 target_originator = 1;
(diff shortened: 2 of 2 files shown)


proto/xmtpv4/message_api/message_api.proto
```diff
@@ -5,20 +5,32 @@ package xmtp.xmtpv4;
...
+import "mls/api/v1/mls.proto";
 
 option go_package = "github.com/xmtp/proto/v3/go/xmtpv4/message_api";
 
-// Authenticated data within the MlsPrivateMessage
+// Data visible to the server that has been signed by the client.
+// Included in one of:
+// - MLS PrivateMessage.authenticated_data for messages and commits
+// - MLS KeyPackageTBS.extensions for key packages
+// - Omitted for Welcome messages
```
richardhuaaa, line 16: I chose to leverage existing fields rather than creating an extra client signature. A drawback is that I can't find an equivalent for Welcome messages, but fortunately causal ordering doesn't make sense for welcome messages (see my comment on the XIP draft).

If we added a standardized signature field for all message types (or just a piecemeal one created for Welcome messages specifically), we could also move the `target_topic` below into the `AuthenticatedData`. If we were to do this, we would need to be careful that the signature is not visible to the server, or we break anonymity.

For now, I assume it's not worth it - the current approach allows us to not modify client interfaces too much, and avoid extra signatures.

proto/xmtpv4/message_api/message_api.proto
```diff
@@ -5,20 +5,32 @@ package xmtp.xmtpv4;
...
-  bytes data = 1;
-  bytes sender_hmac = 2;
+  uint32 target_originator = 1;
+  bytes target_topic = 2;
+  oneof payload {
+    xmtp.mls.api.v1.GroupMessageInput group_message = 3;
+    xmtp.mls.api.v1.WelcomeMessageInput welcome_message = 4;
+    xmtp.mls.api.v1.RegisterInstallationRequest register_installation = 5;
+    xmtp.mls.api.v1.UploadKeyPackageRequest upload_key_package = 6;
+    xmtp.mls.api.v1.RevokeInstallationRequest revoke_installation = 7;
```
richardhuaaa, line 32: Reusing all of the old API inputs minimizes client changes, and also allows us to port over any of the old server validation logic, if we choose to keep it. We can also use this field to determine what goes onto blockchain vs replication sub-systems. Note the server is not *forced* to do any custom logic for each type here - it can choose to just serialize all of this and store it in the DB as a byte array.

One difference from the old API will be the outputs from the query API's, which are a slightly different shape. The old API had many different output types, which were mostly equivalent to the input types, with a sequence ID and timestamp added. Our `GatewayEnvelope` basically includes the same metadata on a single type.

I'm hoping that on the client, we can reflect these changes as a thin wrapper on the API client.

.protolint.yaml
```diff
@@ -44,6 +44,10 @@ lint:
       - ENUM_FIELDS_HAVE_COMMENT
       # This one doesn't like single word RPCs, which I have no problem with
       - RPC_NAMES_CASE
+      # In practice this creates a lot of meaningless comments, we can trust
+      # authors to determine when comments will be useful or not
+      - MESSAGES_HAVE_COMMENT
```
neekolas, line 49: Can't argue with the fact that the comments this lint rule has encouraged are basically useless.

proto/xmtpv4/message_api/message_api.proto
neekolas: To access those fields wouldn't the server need to be able to read the TLS serialized data?

proto/xmtpv4/message_api/message_api.proto
neekolas: That seems like a sensible strategy. I wasn't the one pushing for having all different types of APIs for reads/writes, and now that I've been living with it for 6 months I'm not sure it adds much value.

proto/xmtpv4/message_api/message_api.proto
```diff
@@ -109,6 +123,26 @@ message QueryEnvelopesResponse {
...
 
+message PublishEnvelopeRequest {
+  PayerEnvelope payer_envelope = 1;
+}
+
+message PublishEnvelopeResponse {
+  message PublishEnvelopeSuccess {
+    OriginatorEnvelope originator_envelope = 1;
+  }
+  message PublishEnvelopeFailure {
```
neekolas, line 134: Can't we just handle the error cases with GRPC errors and metadata outlining the reason?

Then all the automatically generated metrics are meaningful, and when we have a GRPC gateway endpoint we also get HTTP status codes correctly set.

richardhuaaa: Totally fair, I misremembered the outcome of our previous discussion. Will just remove it

proto/xmtpv4/message_api/message_api.proto
richardhuaaa: It's a good point. I don't see an existing library for doing the TLS deserialization in golang, let me know if you find anything. It seems the options are:

1. Expose a callable Rust library to the server for extracting the `AuthenticatedData` from the TLS serialized payload.
2. Add the `AuthenticatedData` as a plaintext field on the `ClientEnvelope`, add a client signature somewhere inside the encrypted payload, and add client logic to validate it after decryption.

(1) is a bit more complex, (2) is more duplicative. Do you have any thoughts here?

neekolas (approved):

proto/xmtpv4/message_api/message_api.proto
richardhuaaa: Summary of discussion: we are going to reuse the validation service (callable Rust library)

PR #201, opened 2024-08-30
Cursor format that allows clients to change gateways

richardhuaaa:
1. Define a 'vector clock' type with clearly defined rules.
2. Allow a vector clock to be specified as the cursor, for clients that wish to change gateway nodes. Clients will be tracking this information regardless, and this allows us to avoid any heuristics such as a 'lookback period'. The client's first query to the node will specify a vector clock as the cursor, with subsequent queries free to fall back to using the gateway SID again.

A consideration to be made here is whether we should dispense with gateway sids completely, and use the vector clock for all queries. What holds me back is whether we need to worry about the additional message size for queries, and whether the SQL query will be slower (it'll have one `OR originator_node_id = ... AND originator_sequence_id > ...` for each entry in the vector clock). Getting rid of gateway sids would simplify a lot though.

### DIFF
--- a/proto/xmtpv4/message_api/message_api.proto
+++ b/proto/xmtpv4/message_api/message_api.proto
@@ -9,11 +9,17 @@ import "mls/api/v1/mls.proto";
 
 option go_package = "github.com/xmtp/proto/v3/go/xmtpv4/message_api";
 
+// The last seen entry per originator. Originators that have not been seen are omitted.
+// Entries MUST be sorted in ascending order, so that smaller node ID's appear first.
+message VectorClock {
+  repeated uint64 originator_sids = 1;
+}
+
 // Data visible to the server that has been authenticated by the client.
 message AuthenticatedData {
   uint32 target_originator = 1;
   bytes target_topic = 2;
-  repeated uint64 last_originator_sids = 3;
+  VectorClock last_seen = 3;
 }
 
 message ClientEnvelope {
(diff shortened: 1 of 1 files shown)


neekolas:
The size of the query payloads doesn't seem that bad to me. For most groups it's likely only a handful of nodes X 64 bytes. A few hundred bytes per page even in groups spread across many nodes. 

These compound queries are going to be slower, but I don't think we can rule this out without clear data showing the perf is unacceptable. Reads can be run against replicas, so reads are inherently scalable. 

If we were to do this I'd be tempted to go all the way and remove originator sequence IDs altogether. That's where we get the maximum simplification and benefits.

richardhuaaa:
> If we were to do this I'd be tempted to go all the way and remove originator sequence IDs altogether

Just double checking, you mean gateway sequence IDs, right?

neekolas:
> you mean gateway sequence IDs, right?

Yes, sorry

Thinking about this a little more, I wonder if vector clocks are actually going to be bigger than I first thought. To completely check a group topic you can't just provide the clocks for the gateways who have originated messages in that group. A new message could come from any gateway. It could even come from gateways that are currently unhealthy, but previously were healthy and received messages. 

Does this mean that we really need to include every node (past and present) in the vector clock to do this kind of query? That might be an unsustainable number.

richardhuaaa:
Messages can come from new originators, but you don't need to specify those originators in the clock to achieve this (unspecified originators are assumed to be at cursor 0). The query on the server can assume that too, as follows -

```
...
WHERE
   ...
   AND (originator_node_id NOT IN [originator_1, originator_2]
     OR (originator_node_id = originator_1 AND originator_sequence_id > originator_1_last_seen)
     OR (originator_node_id = originator_2 AND originator_sequence_id > originator_2_last_seen));
```

neekolas:
That's a good point

neekolas (approved):

PR #202, opened 2024-09-03
Remove defunct protos

richardhuaaa:
There is a bunch of defunct code on both client and server I want to clean up - this should be helpful for new hires, backwards-compatibility for the replication project, debugging, and auditability. Starting with the protos first


proto/mls_validation/v1/service.proto
```diff
@@ -9,40 +9,19 @@ option go_package = "github.com/xmtp/proto/v3/go/mls_validation/v1";
...
   rpc GetAssociationState(GetAssociationStateRequest) returns (GetAssociationStateResponse) {}
-  
+
   // Validates InboxID key packages and returns credential information for them, without checking
   // whether an InboxId <> InstallationPublicKey pair is really valid.
-  rpc ValidateInboxIdKeyPackages(ValidateKeyPackagesRequest) returns (ValidateInboxIdKeyPackagesResponse) {}
-  
-  // Validate an InboxID Key Package
-  // need public key possibly
-  rpc ValidateInboxIds(ValidateInboxIdsRequest) returns (ValidateInboxIdsResponse) {}
```
richardhuaaa, line 27: This one is one of the newer added endpoints, but I *think* it's unused based on my grepping. Would love to confirm

insipx: I think this might still be used in `register_installation`? https://github.com/xmtp/xmtp-node-go/blob/987ef538b1c6ffbd746d2a20dccfed2ec035b595/pkg/mls/api/v1/service.go#L126

it was part of the [validation pr](https://github.com/xmtp/xmtp-node-go/pull/389)

Although, I'm not sure we're using any of the `ValidateKeyPackages` methods in `mls_validation_service` -- would love to get rid of those in `libxmtp` + the `verified_key_package` file + clean up validation service if possible so we don't have two versions of it.

@neekolas work with updating to register a keypackage/installation at the same time may have changed this though, but it seems like we're still using at least the `ValidateInboxId` code in `mls_validation_service` + for installation registration

richardhuaaa: I think the `RegisterInstallation` endpoint is also defunct, I've removed it in this PR. Is there any other location I maybe missed?

richardhuaaa: And yep, I removed `ValidateKeyPackages` above, that's the precursor to the client changes you've mentioned here

insipx: yeahp LGTM going to be nice to remove this code :)

richardhuaaa: I guess it's difficult to verify because we have a bunch of defunct code that depends on each other - if we're unsure, I'm fine with going ahead with removing it here, and then reverting it if the cascading code changes on client/server reveal that it's still needed

neekolas (commented):
Thanks for tackling. Think this checks off a few boxes in Phase 2 of [this issue](https://github.com/xmtp/xmtp-node-go/issues/399)

proto/mls_validation/v1/service.proto
neekolas: Removing dead code from the client and protos is safe. Then once that is removed we just need to make sure that we're removing the same thing from the server. And that all clients have updated to a version that doesn't call those endpoints (`RegisterInstallation` is the only one a somewhat recent client version would call).

neekolas (approved):

PR #207, opened 2024-09-16
AssociationState created_at

codabrink:


proto/identity/associations/association.proto
```diff
@@ -91,6 +91,7 @@ message AssociationState {
   repeated MemberMap members = 2;
   string recovery_address = 3;
   repeated bytes seen_signatures = 4;
+  uint64 created_at = 5;
```
neekolas, line 94: We'd want to have the created_at field on each member in the `MemberMap`

neekolas (approved):

proto/identity/associations/association.proto
codabrink: Got it

insipx (approved):

PR #210, opened 2024-09-19
Bidirectional BatchSubscribeEnvelopes stream

richardhuaaa:
Gives clients the ability to modify what they are subscribed to without tearing down and re-establishing the stream.

This is a little different from the old server - the client stream consists of deltas to the subscription request, rather than replacing the subscription request entirely. This saves the server implementation from having to tear down and re-establish the stream, and should be a little easier to implement.

https://github.com/xmtp/xmtpd/issues/167


neekolas (approved):

richardhuaaa:
We decided this is no longer necessary - clients will stream by cursor, and have only one cursor per stream, with an arbitrary number of topics

PR #211, opened 2024-09-26
SCW Verification

codabrink:


proto/mls/message_contents/association.proto
```diff
@@ -0,0 +1,54 @@
+// Associations and signatures
```
neekolas, line 1: I think we might be able to leave this file deleted

proto/mls/message_contents/credential.proto
```diff
@@ -0,0 +1,31 @@
+// Credentials and revocations
```
neekolas, line 1: I _believe_ we can leave this deleted too

proto/mls/message_contents/association.proto
codabrink: Cool

proto/mls/message_contents/credential.proto
codabrink: Gone

insipx (approved):

PR #215, opened 2024-10-03
Move scw block number to verification request

codabrink:
Sibling: https://github.com/xmtp/libxmtp/pull/1112

### DIFF
--- a/proto/identity/associations/signature.proto
+++ b/proto/identity/associations/signature.proto
@@ -27,8 +27,6 @@ message SmartContractWalletSignature {
   // CAIP-10 string
   // https://github.com/ChainAgnostic/CAIPs/blob/main/CAIPs/caip-10.md
   string account_id = 1;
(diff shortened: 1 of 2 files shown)


insipx (approved):

proto/identity/associations/signature.proto
```diff
@@ -27,8 +27,6 @@ message SmartContractWalletSignature {
   // CAIP-10 string
   // https://github.com/ChainAgnostic/CAIPs/blob/main/CAIPs/caip-10.md
   string account_id = 1;
-  // Specify the block number to verify the signature against
-  uint64 block_number = 2;
```
neekolas, line 31: We still need the block number in here, since this is what we send to everyone else to verify the signature. We want them to verify it at a specific block

codabrink: Yeah, I realized that, was just about to move this back into draft.

PR #223, opened 2024-10-18
Payer API Refactor

neekolas:
## tl;dr

- Creates a new payer API
- Breaks envelopes into their own package so they can be referenced from both the message api and the payer API
- Renames `PublishEnvelopes` to `PublishPaidEnvelopes` and adds a new `PublishUnpaidEnvelopes` API


proto/xmtpv4/message_api/message_api.proto
```diff
@@ -159,9 +102,9 @@ service ReplicationApi {
   }
 
   // Publish envelope
-  rpc PublishEnvelopes(PublishEnvelopesRequest) returns (PublishEnvelopesResponse) {
+  rpc PublishPaidEnvelopes(PublishPaidEnvelopesRequest) returns (PublishPaidEnvelopesResponse) {
```
neekolas, line 105: I didn't want to have two APIs named `PublishEnvelopes` that expect different request/response types. That feels like a footgun.

This is the best name I could come up with.

mkysel (approved):
looks straight forward enough!

proto/xmtpv4/payer_api/payer_api.proto
```diff
@@ -0,0 +1,28 @@
...
+message PublishUnpaidEnvelopesResponse {
+  repeated xmtp.xmtpv4.envelopes.OriginatorEnvelope originator_envelopes = 1;
+}
+
+// A narrowly scoped API for publishing messages through a payer
+service PayerApi {
+  // Publish envelope
+  rpc PublishUnpaidEnvelopes(PublishUnpaidEnvelopesRequest) returns (PublishUnpaidEnvelopesResponse) {
+    option (google.api.http) = {
+      post: "/mls/v2/payer/publish-envelopes"
```
mkysel, line 24: is this what we expect libxmtp to use?

neekolas: The `google.api.http` is for GRPC Gateway (which turns our GRPC Proto API into a HTTP/JSON API).

It would only be used in libxmtp in WASM environments, which don't have GRPC support

proto/xmtpv4/message_api/message_api.proto
richardhuaaa: I'm wondering if `PublishPayerEnvelopes` and `PublishClientEnvelopes` would be more consistent?

richardhuaaa (approved):

neekolas:
@richardhuaaa 
> I'm wondering if PublishPayerEnvelopes and PublishClientEnvelopes would be more consistent?

That was my #2 choice, but I don't really have a strong preference. Happy to switch it over.

PR #224, opened 2024-10-18
Remove dependency on originator node for blockchain envelopes

richardhuaaa:
We allow anyone to publish to the blockchain directly, so there is no need to specify the `publisher_node_id`. The `originator_ns` field on the `UnsignedOriginatorEnvelope` will be derived from the block time.

### DIFF
--- a/proto/xmtpv4/message_api/message_api.proto
+++ b/proto/xmtpv4/message_api/message_api.proto
@@ -41,8 +41,7 @@ message PayerEnvelope {
   xmtp.identity.associations.RecoverableEcdsaSignature payer_signature = 2;
 }
 
-// For blockchain envelopes, the originator_sid is set by the smart contract,
-// but the originator_ns is set by the publishing node
+// For blockchain envelopes, these fields are set by the smart contract
 message UnsignedOriginatorEnvelope {
   uint32 originator_node_id = 1;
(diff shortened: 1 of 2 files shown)


neekolas (approved):

proto/xmtpv4/message_api/message_api.proto
```diff
@@ -53,7 +52,6 @@ message UnsignedOriginatorEnvelope {
 // An alternative to a signature for blockchain payloads
 message BlockchainProof {
   uint64 block_number = 1;
-  uint32 publisher_node_id = 2;
```
neekolas, line 56: We could also include the transaction hash, since this proof gets generated when the message is read off the chain. That would make it easier to verify

richardhuaaa: Do we expect recipients to verify both the `block_number` and `transaction_hash`? Or would the `transaction_hash` alone be sufficient to find the proof? Just trying to avoid redundant fields that need to be kept consistent where possible

richardhuaaa: (Also, you can merge first, I'll apply this on top of your change)

neekolas: Think the hash could replace the block number, since you can get the block number with the hash

PR #235, opened 2024-12-18
Remove VectorClock type

richardhuaaa:
We are no longer referring to this as a 'vector clock', instead it could be more accurately called a cursor.

Seeing as we are renaming anyway, I thought it might be a good opportunity to unnest it, as it complicated the code for using the field. In the original layout, the field could be referred to by three nested names: `last_seen`, `VectorClock`, `node_id_to_sequence_id`.

The issue though is that this is a breaking change, and would require resetting the DB. We may have to do this not just for our own instance of the node, but in any other partner nodes. If this ends up being too difficult, I'm happy to avoid the unnesting and simply rename `VectorClock`->`Cursor`. Although it could be a good thing to run through a breaking change as practice regardless.

### DIFF
--- a/proto/identity/api/v1/identity.proto
+++ b/proto/identity/api/v1/identity.proto
@@ -4,16 +4,15 @@ package xmtp.identity.api.v1;
 
 import "google/api/annotations.proto";
 import "identity/associations/association.proto";
-import "identity/associations/signature.proto";
 import "protoc-gen-openapiv2/options/annotations.proto";
 
 option go_package = "github.com/xmtp/proto/v3/go/mls/api/v1";
 option java_package = "org.xmtp.proto.mls.api.v1";
 option (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_swagger) = {
   info: {
-    title: "IdentityApi";
-    version: "1.0";
-  };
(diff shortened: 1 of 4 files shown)


proto/identity/api/v1/identity.proto
```diff
@@ -4,16 +4,15 @@ package xmtp.identity.api.v1;
 
 import "google/api/annotations.proto";
 import "identity/associations/association.proto";
-import "identity/associations/signature.proto";
```
richardhuaaa, line 7: Unrelated lint/auto-format changes

neekolas (commented):
I think we should do a trial run of a breaking change.

I don't think the week before Christmas is the week to do it. Very soon it's going to be hard to get a hold of node operators. If you want this update before January, I'd suggest doing it in the non-breaking way.

richardhuaaa:
There is no rush on this - just put this up so that we have it on our minds. Can wait until January

richardhuaaa:
Okay, I figure it is easier to separate the breaking change to a separate PR if we really need it. Just renaming for now which can unblock us on more short-term things like releasing the XIP

neekolas (approved):

PR #237, opened 2024-12-20
Flesh out misbehavior reports

richardhuaaa:
1. Fleshes out the misbehavior report types and fields
2. Defines endpoints for storing and retrieving misbehavior reports


proto/xmtpv4/envelopes/envelopes.proto
```diff
@@ -60,3 +60,45 @@ message OriginatorEnvelope {
...
+  LivenessFailure liveness = 4;
+  SafetyFailure safety = 5;
+  // Nodes must verify this field is false for client-submitted reports
+  bool submitted_by_node = 6;
+}
+
+message MisbehaviorReport {
+  // Server time when the report was stored. Used only for querying reports.
+  // This field is not signed.
+  uint64 server_time_ns = 1;
```
richardhuaaa, line 100: We need this separately to the reporter time (which may have client skew), so that the query endpoint can query all payloads larger than a certain timestamp. Could have used a sequence ID or a timestamp, decided to use a timestamp so that it's simpler, while also avoiding confusion with cursors in the main replication protocol

proto/xmtpv4/envelopes/envelopes.proto
```diff
@@ -60,3 +60,45 @@ message OriginatorEnvelope {
...
+  bytes unsigned_misbehavior_report = 2;
+  // Signed by the node hosting the report
+  xmtp.identity.associations.RecoverableEcdsaSignature signature = 3;
```
richardhuaaa, line 103: We have the node's signature here. Alternative approaches:

- No signature - it makes it a bit tricky to handle cases where a node retracts or denies a report it previously made. It also seems useful to know which node client reports came in through, as opposed to them going into one big pool.
- Add a client signature - I think in most cases clients will want to stay anonymous. Adding your public key to a set of group payloads may end up deanonymizing a group.

When we define incentives for submitting reports, some entities may want to be rewarded. When we get there, perhaps we could add a `claimant_public_key` field to the `UnsignedMisbehaviorReport`. Probably don't need an additional signature for that.

proto/xmtpv4/message_api/message_api.proto
```diff
@@ -83,37 +68,60 @@ message GetInboxIdsResponse {
...
 
-// Replication API
+message SubmitMisbehaviorReportRequest {
+  xmtp.xmtpv4.envelopes.UnsignedMisbehaviorReport report = 1;
+}
+
+message SubmitMisbehaviorReportResponse {}
+
+message QueryMisbehaviorReportsRequest {
+  uint64 after_ns = 1;
```
richardhuaaa, line 78: I think this is the minimum set of parameters we should be able to query by, but in the future it may make sense to add queries by `misbehaving_node_id` or misbehavior type or queries that exclude client reports

proto/xmtpv4/message_api/misbehavior_api.proto
```diff
@@ -0,0 +1,83 @@
...
+
+message SafetyFailure {
+  repeated xmtp.xmtpv4.envelopes.OriginatorEnvelope envelopes = 1;
+}
+
+message UnsignedMisbehaviorReport {
+  uint64 reporter_time_ns = 1;
+  uint32 misbehaving_node_id = 2;
+  Misbehavior type = 3;
+  LivenessFailure liveness = 4;
```
neekolas, line 40: We don't want LivenessFailure and SafetyFailure to be a `oneof`?

richardhuaaa: In most languages, `oneof` becomes a nested field. I suppose it's a question of developer ergonomics vs documenting expectations clearly, seeing as fields are optional either way. I'll go ahead and make it a `oneof`

neekolas (approved):

PR #238, opened 2024-12-30
feat: add SubscribeAssociationChanges

neekolas:
## tl;dr

- Creates protos for streaming association state changes

## Issues

https://github.com/xmtp/xmtp-node-go/issues/410

### DIFF
--- a/proto/identity/api/v1/identity.proto
+++ b/proto/identity/api/v1/identity.proto
@@ -155,9 +155,16 @@ message SubscribeAssociationChangesResponse {
     string inbox_id = 2;
   }
 
+  // A revocation of an account_address -> inbox_id association
+  message AccountAddressRevocation {
+    string account_address = 1;
+    string inbox_id = 2;
+  }
+
   uint64 timestamp_ns = 1;
 
   oneof change {
     AccountAddressAssociation account_address_association = 2;
+    AccountAddressRevocation account_address_revocation = 3;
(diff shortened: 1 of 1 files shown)


proto/identity/api/v1/identity.proto
```diff
@@ -137,3 +145,27 @@ message GetInboxIdsResponse {
 
   repeated Response responses = 1;
 }
+
+// Request to subscribe to association changes, triggered by Identity Updates
+message SubscribeAssociationChangesRequest {}
+
+message SubscribeAssociationChangesResponse {
```
richardhuaaa, line 152: Is there any world where this can just be a stream of identity updates, with the parsing of the stream done client-side? Would be simpler/more compatible with decentralization/requiring less server trust

neekolas: I'm hoping in a decentralized world we can create a subgraph that does that aggregation by reading events from the blockchain and running our code. Then we can slice and dice the data any way we want.

But for now, it's a big ask for clients to run `libxmtp` from whatever language/framework they are using and correctly store previous association states and apply updates.

richardhuaaa (approved):

mkysel:
@neekolas is this still relevant?

neekolas:
@mkysel Yes. Waiting on feedback from a partner before I merge it

PR #245, opened 2025-02-11
Identity Migration

codabrink:


proto/identity/associations/association.proto
```diff
@@ -11,11 +11,25 @@ option java_package = "org.xmtp.proto.identity.associations";
...
 
+// Passkey identifier
+message Passkey {
+    bytes key = 1;
+    optional string relying_partner = 2;
+}
+
+// List of identity kinds
+enum IdentifierKind {
+  IDENTIFIER_KIND_UNSPECIFIED = 0; // Ethereum on old clients
```
mchenani, line 28: Q: for old clients, can we check if their identity is eth to set them as `IDENTIFIER_KIND_ETHEREUM`?

mchenani (approved):

insipx (approved):

PR #250, opened 2025-03-05
fix: add should push to the group messages

nplasterer:


proto/mls/api/v1/mls.proto
```diff
@@ -163,6 +164,7 @@ message GroupMessageInput {
   message V1 {
     bytes data = 1; // Serialized MlsProtocolMessage
     bytes sender_hmac = 2;
+    optional bool should_push = 4;
```
nplasterer, line 167: Does making this optional matter? Old messages sent on the old format will just not have this option set.

cameronvoell: this looks okay to me

neekolas: What is the difference between leaving this option unset and setting it to true?

Seems like the push server would treat them the same

cameronvoell, line 167:
```suggestion
    optional bool should_push = 3;
```

cameronvoell: should be 3 though not 4

proto/mls/api/v1/mls.proto
```diff
@@ -150,6 +150,7 @@ message GroupMessage {
     bytes group_id = 3;
     bytes data = 4;
     bytes sender_hmac = 5;
+    optional bool should_push = 4;
```
neekolas, line 153: This has a duplicate field number

cameronvoell (approved):
no need to ad a V2 since we're not removing / reo-ordering, changing types of existing. looks good once you update that second one from 3 => 4

PR #254, opened 2025-03-20
Remove V2 Protos

nplasterer:
Lets clean up some of these V2 protos


proto/message_contents/signature.proto
nplasterer, line 1: Is this being used at all in V3? I'm assuming not

insipx (approved):

PR #262, opened 2025-04-29
Device Sync Content

codabrink:
### Define Protocol Buffer message structures for device synchronization content in device sync module
* Introduces new Protocol Buffer definitions in [`content.proto`](https://github.com/xmtp/proto/pull/262/files#diff-f9c3d5fbe18115b88a8af6d6419c11934cc6de7f22f97a82a4f0e5948413d51a) establishing message structures for device sync operations including request handling, acknowledgments, and preference updates
* Updates [`protolint.yaml`](https://github.com/xmtp/proto/pull/262/files#diff-78a43d8354af1af40269e37f7a492166c8f58bda146af673b2d99211ae6fdfef) to enable `repeated_field_names_pluralized` rule with exceptions for `admin_list` and `super_admin_list` fields

#### 📍Where to Start
Start with the new [`DeviceSyncContent`](https://github.com/xmtp/proto/pull/262/files#diff-f9c3d5fbe18115b88a8af6d6419c11934cc6de7f22f97a82a4f0e5948413d51a) message definition which serves as the main container for all device sync operations.

----

_[Macroscope](https://app.macroscope.com) summarized e946956._

### DIFF
--- a/proto/device_sync/content.proto
+++ b/proto/device_sync/content.proto
@@ -1,8 +1,8 @@
 syntax = "proto3";
 package xmtp.device_sync.content;
 
-import "mls/message_contents/content.proto";
 import "device_sync/consent_backup.proto";
+import "mls/message_contents/content.proto";
 
 message DeviceSyncContent {
   oneof content {
@@ -15,12 +15,8 @@ message DeviceSyncContent {
 
 message AcknowledgeKind {
   oneof kind {
-    SyncGroupPresence sync_group_presence = 1;
-    RequestAcknowledge request = 2;
+    RequestAcknowledge request = 1;
   }
-}
-
-message SyncGroupPresence {
 }
 
 message RequestAcknowledge {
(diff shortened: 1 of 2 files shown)


proto/device_sync/content.proto
```diff
@@ -0,0 +1,43 @@
...
+    RequestAcknowledge request = 2;
+  }
+}
+
+message SyncGroupPresence {
+}
+
+message RequestAcknowledge {
+  string request_id = 1;
+}
```
richardhuaaa, line 28: Would it make sense to collapse this all into a single `DeviceSyncAcknowledge` message with an optional `request_id` (or a `request_id` of 0 if it was an implicit request)?

If we're using conversation stitching (abstracting multiple sync groups so that we can pretend there is a single long-lived sync group), I'm also wondering if we could do away with implicit requests and have all requests be explicit?

PR #264, opened 2025-05-06
Add GetReaderNode to payer api

fbac:
Initial stages of the work around https://github.com/xmtp/libxmtp/issues/1547

The Payer API will expose a `GetReaderNode` method, for clients to retrieve a node URL to read from.

The rpc itself is defined as the bare minimum and will return a randomized node from the pool of healthy nodes. It could be easily extended to include more fields in the future, such as:

```
   message GetReaderNodeRequest {
     optional string region_preference = 1;
     optional bool high_availability = 2;
   }

   message GetReaderNodeResponse {
     string reader_node_url = 1;
     repeated string backup_node_urls = 2;
     optional int64 ttl_seconds = 3;
   }
```


mkysel (approved):
reasonable. looks like the right level of abstraction

PR #272, opened 2025-06-04
Add cursor to welcome

codabrink:


proto/mls/api/v1/mls.proto
```diff
@@ -137,6 +138,7 @@ message WelcomeMessageInput {
     bytes data = 2;
     bytes hpke_public_key = 3;
     xmtp.mls.message_contents.WelcomeWrapperAlgorithm wrapper_algorithm = 4;
+    uint64 group_refresh_state_cursor = 5;
```
mchenani, line 141: Wryt of something like:
message_cursor_at_join
Or just
message_cursor

codabrink: The value comes from the sender's refresh_state group cursor value. I like the `group_refresh_state_cursor` as it describes where the value came from. Correct me if I'm wrong - but if a welcome doesn't immediately succeed, this cursor value could be off by 1 or 2 messages, right?

neekolas: Why not set it at the exact id of the commit?

richardhuaaa (approved):

richardhuaaa:
I'm not sure if the message cursor is authenticated, and/or if the server can tamper with it. But given the tight deadline we can revisit this later

PR #282, opened 2025-07-16
feat: remove is-commit - will be determined from payload

mkysel:
### Remove `is_commit` field from `AuthenticatedData` message in protocol buffer definition as it will be determined from payload
The `is_commit` boolean field (tag 4) is removed from the `AuthenticatedData` message definition in [proto/xmtpv4/envelopes/envelopes.proto](https://github.com/xmtp/proto/pull/282/files#diff-f812a33b13a82e881082c50bb3818e6aad1d114d1141279972acbc496a645fda). A comment is added indicating that tag 4 should not be reused, following the same pattern as the existing comment for tag 1 which was previously used by `target_originator`.

#### 📍Where to Start
Start with the `AuthenticatedData` message definition in [proto/xmtpv4/envelopes/envelopes.proto](https://github.com/xmtp/proto/pull/282/files#diff-f812a33b13a82e881082c50bb3818e6aad1d114d1141279972acbc496a645fda).

----

_[Macroscope](https://app.macroscope.com) summarized 4007e6c._

### DIFF
--- a/proto/xmtpv4/envelopes/envelopes.proto
+++ b/proto/xmtpv4/envelopes/envelopes.proto
@@ -20,7 +20,7 @@ message AuthenticatedData {
   // Do NOT reuse tag 1 — previously used by target_originator
   bytes target_topic = 2;
   Cursor depends_on = 3;
-  // Do NOT reuse tag 1 — previously used by is_commit
+  // Do NOT reuse tag 4 — previously used by is_commit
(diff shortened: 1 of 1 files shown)


mkysel:
the go failure is expected. It's a chicken&egg problem.

fbac (approved):

PR #288, opened 2025-08-19
feat: readd request and intent

richardhuaaa:
Changes:
1. Defines a 'oneshot' conversation type
2. Defines a 'readd request' oneshot message, which is used to request fork recovery
3. Defines a readd intent, which is the response to the request. The creator of the intent will publish a commit and then send welcomes to the requester.

__________

Optional notes:
Readd requests need to be sent to superadmins outside of the main group (as the sender is forked), however they still need MLS level authentication and privacy guarantees.

We could create a single-use group and then send a message on it, but given only a single message is needed, an alternative is to embed the message on the immutable metadata of the group itself, and have the recipient 'receive' the message at the time of decrypting the welcome, without persisting the group (ie: use openMLS to construct the group/StagedWelcome in-memory, pull the message from the immutable metadata, then throw the group away). That would allow us to avoid:
- Delays from the sender needing to send an additional message on the group after the welcomes have been sent
- Delays from the recipient needing to sync the group after receiving the welcome to receive the message
- Any unintentional consequences around group storage, syncing on it in the future, calling maybe_update_installations, or rendering it in UI. Also means we don't create a new topic on the server.

Are there traps that I'm falling into here? And please let me know if a better name exists!

### DIFF
--- a/proto/mls/message_contents/content.proto
+++ b/proto/mls/message_contents/content.proto
@@ -5,7 +5,6 @@ syntax = "proto3";
 package xmtp.mls.message_contents;
 
 import "device_sync/content.proto";
-import "mls/message_contents/out_of_band.proto";
 
 option go_package = "github.com/xmtp/proto/v3/go/mls/message_contents";
 option java_package = "org.xmtp.proto.mls.message.contents";
@@ -70,9 +69,11 @@ message PlaintextEnvelope {
       xmtp.device_sync.content.DeviceSyncReply device_sync_reply = 4;
       // A serialized user preference update
       xmtp.device_sync.content.V1UserPreferenceUpdate user_preference_update = 5;
-      // A readd request for fork recovery
-      ReaddRequest readd_request = 6;
     }
+
+    // Removed; moved to oneshot message
+    reserved 6;
+    reserved "readd_request";
   }
 
   // Selector which declares which version of the EncodedContent this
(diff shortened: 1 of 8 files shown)


proto/mls/message_contents/group_mutable_metadata.proto
```diff
@@ -14,9 +14,9 @@ message GroupMutableMetadataV1 {
   // Creator starts as only super_admin
   // Only super_admin can add/remove other super_admin
   Inboxes super_admin_list = 3;
-  // The ED25519 private key used to sign commit log entries
-  // Must match the first entry in the commit log to be valid
-  optional bytes commit_log_signer = 4;
```
richardhuaaa, line 19: This is not related to oneshot messages - removed as this is not being used in any live client

proto/mls/message_contents/content.proto
```diff
@@ -70,8 +70,9 @@ message PlaintextEnvelope {
       xmtp.device_sync.content.DeviceSyncReply device_sync_reply = 4;
       // A serialized user preference update
       xmtp.device_sync.content.V1UserPreferenceUpdate user_preference_update = 5;
-      // A readd request for fork recovery
-      ReaddRequest readd_request = 6;
+      // Removed; moved to oneshot message
+      reserved 6;
+      reserved "readd_request";
```
richardhuaaa, line 75: Not being used in any live client

richardhuaaa:
One question here is how this group type is handled on old clients. That might be a little hairy

cameronvoell:
@richardhuaaa seems to be failing a test for creating golang code in xmtpd, mind taking a look?

PR #296, opened 2025-09-29
Upgrade deps, refactor build process

rygine:


ts/node.ts
```diff
@@ -1,44 +0,0 @@
-function getGlobal(): any | undefined {
```
neekolas, line 1: I don't believe this is needed anymore now that

1. Node has improved its `fetch` support a lot
2. We use GRPC web

Should be safe to remove.

rygine:
closing this in favor of #307

PR #333, opened 2026-05-21
feat: AppDataUpdateData intent serialization proto

tylerhawkes:
## Summary

Adds `AppDataUpdateData` + `AppDataUpdateDeltaWithBase` to `xmtp.mls.database` for local-only intent serialization of libxmtp's new generic `IntentKind::AppDataUpdate` (https://github.com/xmtp/libxmtp/pull/3664).

The proto carries:
- `component_id: uint32` (libxmtp's `u16` widened — proto has no u16)
- `oneof op` with:
  - `bytes replace` — full-replace components (Bytes/String typed)
  - `AppDataUpdateDeltaWithBase delta_with_base` — collection components (TlsMap/TlsSet) with `pre_value`+`post_value` for 3-way merge

Never crosses the MLS wire; stored only in libxmtp's local SQLite `group_intents` table.

## Test plan

- [x] Pinned at SHA `7b8260c3041389bf706fb665460adbbd520d2be7`.
- [x] libxmtp regen succeeds with `GEN_PROTOS=true cargo check -p xmtp_proto`.
- [x] Round-trip + edge-case tests in libxmtp PR #3664 pass.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

> [!NOTE]
> ### Add `AppDataUpdateData` protobuf message for app data update intent serialization
> Adds a new versioned protobuf message `AppDataUpdateData` in [intents.proto](https://github.com/xmtp/proto/pull/333/files#diff-6f0fe39046ec77adf0be4b95d56a75c664665c04f7c30965c66d2e5cd24319f0) with a `V1` envelope containing `component_id` (uint32) and `payload` (bytes). Inline comments document the expected payload semantics for `Bytes`/`String`, `TlsMap`, and `TlsSet` component types.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized de79332.</sup>
> 
>

### DIFF
--- a/proto/mls/database/intents.proto
+++ b/proto/mls/database/intents.proto
@@ -165,6 +165,48 @@ message UpdatePermissionData {
   }
 }
 
+// Generic AppData write intent — local-only serialization for the
+// `group_intents` table. Replaces the proliferation of per-component
+// IntentKind/IntentData pairs with a single shape that carries the
+// target component_id and the same payload bytes that will be written
+// to the on-wire `AppDataUpdate` proposal. Interpretation of `payload`
+// is determined by the target component's `ComponentType` in the
+// registry:
+//
+//   * `Bytes` / `String` typed components — payload is the new value
+//     written verbatim (last-writer-wins).
+//   * `TlsMap` typed components — payload is a TLS-encoded
+//     `TlsMapDelta<K, V>` carrying `Insert` / `Update` / `Delete`
+//     mutations. Apply semantics on the receive side are total
+//     (Insert is no-op-if-present, Update is upsert, Delete is
+//     idempotent); the strict variants of the in-process TlsMap API
+//     are not exposed on the wire. No "conflict" failure path.
+//   * `TlsSet` typed components — payload is a TLS-encoded
+//     `TlsSetDelta<E>` (`Add` is no-op-if-present, `Remove` is
+//     idempotent).
+//
+// Never appears on the MLS wire; lives only in the local intents
+// table.
+message AppDataUpdateData {
+  // v1 payload shape.
+  message V1 {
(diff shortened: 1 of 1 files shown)


proto/mls/database/intents.proto
```diff
@@ -165,6 +165,45 @@ message UpdatePermissionData {
...
+    // time.
+    AppDataUpdateDeltaWithBase delta_with_base = 3;
+  }
+}
+
+message AppDataUpdateDeltaWithBase {
+  // Full prior bytes of the component as observed when the caller
+  // built the intent. Merge base for the 3-way merge.
+  bytes pre_value = 1;
+  // Full desired post bytes of the component.
```
neekolas: Do you have some reference on how we resolve the residual mutation and do the 3-way-merge?

I like this approach in general, just want to learn more about the process since it seems important to grok moving forward.

neekolas (approved):

proto/mls/database/intents.proto
tylerhawkes: For components that are Maps/Sets you can just write out the current state and the desired state and when doing the commit we can figure out what the delta is to get into the desired state. So if Alix wants to add inboxes x and y, but Bo adds X in the meantime, we can still figure out that we need to add y and just make that change. I'll have to do more tests with interim removals, but the idea is that we can state the end state (either total, or the diffs) and then choose to not publish some of the changes if they've already been made.

tylerhawkes: After thinking about it, we're going to go with last writer wins for now. We can update this later if we want some better semantics around races, but those should be rare enough that we maybe don't care about them.

PR #337, opened 2026-06-12
feat: XIP-83 bidirectional Subscribe RPC on MlsApi

tylerhawkes:
Adds the XIP-83 bidirectional, mutable subscription stream to the v3 MLS API: one long-lived `Subscribe(stream SubscribeRequest) returns (stream SubscribeResponse)` RPC.

- **Mutate-in-place subscriptions over kind-prefixed binary topics** (the XIP-49 §3.3.2 representation shared with the d14n envelopes — thanks @insipx): `Mutate{ adds: [{topic, id_cursor}], removes: [topics] }`. One shape extends to future streamable kinds (key packages, identity updates) with no proto changes; v3 keeps its single `uint64` cursor (`0` = from the beginning; joiners SHOULD seed from the welcome's encrypted `WelcomeMetadata.message_cursor`). No reconnect on membership change.
- **Correlated catch-up waves**: each adding Mutate carries a client-chosen `mutate_id`, echoed on that wave's `CatchupComplete`, so completions stay attributable when waves overlap.
- **Dedicated lifecycle arms** `Started{keepalive_interval_ms, capabilities}` / `CatchupComplete{mutate_id}` instead of a status enum with side fields — frame metadata is unambiguous by construction (no proto3 presence games).
- **Live-boundary markers**: `TopicsLive{topics}` after each topic's history.
- **`history_only` Mutates**: bounded catch-up with no live registration — combined with client half-close this is the stream-native sync flow (server drains the wave, then closes).
- **WebSocket-style liveness**: top-level `Ping`/`Pong`; version pinned per stream (V1 requests ⇒ V1 responses only).

Spec: xmtp/XIPs#139. Server implementation: xmtp/xmtp-node-go#562. Client: xmtp/libxmtp#3769–#3772.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

> [!NOTE]
> ### Add bidirectional `Subscribe` RPC to `MlsApi` for mutable topic subscriptions
> - Adds a new bidirectional streaming RPC `MlsApi.Subscribe` in [mls.proto](https://github.com/xmtp/proto/pull/337/files#diff-ecc5f7ce37e02e9997785d5b1a63c97a03b96d475ba95a22a39866591a22b051) that accepts a stream of `SubscribeRequest` and returns a stream of `SubscribeResponse`.
> - `SubscribeRequest` supports atomic topic mutations (add/remove with `history_only` mode and `mutate_id` correlation), plus `Ping`/`Pong` liveness messages.
> - `SubscribeResponse` carries batched group and welcome messages, a `Started` handshake with keepalive interval, `TopicsLive` transition events, `CatchupComplete` acknowledgments, and `Ping`/`Pong` liveness.
> - The stream is designed as a single long-lived, mutable subscription that can multiplex group-message and welcome topics.
> - The RPC has no HTTP/grpc-gateway mapping and is gRPC-only.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 3988a8d.</sup>
> 
>

### DIFF
--- a/proto/mls/api/v1/mls.proto
+++ b/proto/mls/api/v1/mls.proto
@@ -114,6 +114,12 @@ service MlsApi {
     };
   }
 
+  // Bidirectional subscription (XIP-83). One long-lived stream the client mutates
+  // in place via add/remove topic deltas, with WebSocket-style liveness ping/pong.
+  // A single stream MAY carry both group-message and welcome topics.
+  // gRPC-only: bidirectional streaming has no HTTP/grpc-gateway mapping.
+  rpc Subscribe(stream SubscribeRequest) returns (stream SubscribeResponse) {}
+
   rpc BatchPublishCommitLog(BatchPublishCommitLogRequest) returns (google.protobuf.Empty) {
     option (google.api.http) = {
       post: "/mls/v1/batch-publish-commit-log"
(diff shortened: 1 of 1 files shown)


proto/mls/api/v1/mls.proto
```diff
@@ -404,6 +410,152 @@ message SubscribeWelcomeMessagesRequest {
...
+      Mutate mutate = 1;
+      Ping ping = 2; // liveness challenge (e.g. probe the link after resuming)
+      Pong pong = 3; // answer to a server Ping
+    }
+
+    // Add and/or remove group and welcome subscriptions in place (applied
+    // atomically per frame). The node derives the topic from the id exactly as
+    // it does for the existing Subscribe{Group,Welcome}Messages RPCs.
+    message Mutate {
+      repeated Group add_groups = 1; // begin delivering these group-message topics
```
insipx: maybe we could re-use the [`topic`](https://github.com/xmtp/proto/blob/e6f640a8994dad1779da0280b15be482c0bb4bb8/proto/xmtpv4/envelopes/envelopes.proto#L21) --- [`Rust type`](https://xmtp.github.io/libxmtp/xmtp_proto/types/struct.Topic.html) strategy from d14n, then we could have just two fields, `topic` (groups/welcomes/key packages/etc) and `operation` add/remove etc. then there is also one of `id_cursor` as well, since id and kind is tied into `Topic` and it would be extensible over new types and the other api objects we may want to stream like kps and identity updates

insipx: we could use the type elsewhere we have topics here too (like in `SubscribeResponse`), it could make the resulting proto types smaller and we already have handling for `Topic` libxmtp side

tylerhawkes: Adopted in the latest push — `Mutate` is now `add: [{topic, id_cursor}]` / `remove: [topics]` over the XIP-49 kind-prefixed binary topics, and `TopicsLive` echoes the same representation, so the v3 and d14n vocabularies converge and `xmtp_proto::types::Topic` works unchanged on the client side. Future kinds (key packages, identity updates) land as a kind byte + a capability advertised on `Started`, with no proto changes.

One deviation from your sketch: I went with separate `add`/`remove` lists rather than a `topic` + `operation` enum, since `id_cursor` is only meaningful on adds and an op enum would hang a dead cursor field on every remove. Happy to flip it if you feel strongly. While reshaping I also added `mutate_id` correlation (echoed on the wave's `CatchupComplete`) and split `Started`/`CatchupComplete` into dedicated response arms.

proto/mls/api/v1/mls.proto
```diff
@@ -404,6 +410,149 @@ message SubscribeWelcomeMessagesRequest {
...
+
+      // A topic to subscribe, with the cursor to resume from.
+      message Subscription {
+        bytes topic = 1;
+        // Deliver ids greater than this; 0 = from the beginning. For a newly
+        // joined group, a client SHOULD seed this from the welcome's encrypted
+        // WelcomeMetadata.message_cursor so a new membership does not refetch
+        // pre-join history it cannot decrypt; for a new installation's welcome
+        // topic, 0 is how pending welcomes are collected.
+        uint64 id_cursor = 2;
```
neekolas, line 477: Do we not want this to be a multi-cursor to make it easier to migrate to V4?

neekolas (approved):

insipx (approved):

proto/mls/api/v1/mls.proto
tylerhawkes: I think that's something that each client can handle since we have a v3/d14n separation already.

insipx (approved):

